PR #213, opened 2024-07-23
212 feature rename sheet name

### LINKED ISSUE #212: [TASK] Rename Sheet name
Under the "more" (three dots) icon button on the sheet tab, there are options to duplicate, export, and delete a sheet. Add the option to rename a sheet. When the Rename menu item is clicked, a dialog should appear allowing users to rename the sheet on submit.
image

protanop1c:


neelneelneel:
So this approach doesn't update any other references (i.e. blocks or other notebooks). I think we should update as well. @johbaxter thoughts?

johbaxter:
> So this approach doesn't update any other references (i.e. blocks or other notebooks). I think we should update as well. @johbaxter thoughts?

Meaning a sheet previously named 'sheet-1' , I change it to 'the-best-sheet'.  

Now that it is renamed, I go through my variables, blocks and other notebooks to change the reference there as well.

If so, yes this does need to get added.  Or we message to user the implications of renaming.  I go with just changing reference for them.

@SPRCoreDump make sense? @neelneelneel inline with your thinking?

protanop1c:
> > So this approach doesn't update any other references (i.e. blocks or other notebooks). I think we should update as well. @johbaxter thoughts?
> 
> Meaning a sheet previously named 'sheet-1' , I change it to 'the-best-sheet'.
> 
> Now that it is renamed, I go through my variables, blocks and other notebooks to change the reference there as well.
> 
> If so, yes this does need to get added. Or we message to user the implications of renaming. I go with just changing reference for them.
> 
> @SPRCoreDump make sense? @neelneelneel inline with your thinking?

Was not aware of the reference issues due to this all being novel to me.
I tried for a while to find a way to directly change the ID of the query directly, but I couldn't find an ActionMessage to pass to state.dispatch{} that would allow me to do such a thing...

Can I get an example of what would go wrong with a duplicate + delete?
Also, is this an issue when duplicating as well?

johbaxter:
> > > So this approach doesn't update any other references (i.e. blocks or other notebooks). I think we should update as well. @johbaxter thoughts?
> > 
> > 
> > Meaning a sheet previously named 'sheet-1' , I change it to 'the-best-sheet'.
> > Now that it is renamed, I go through my variables, blocks and other notebooks to change the reference there as well.
> > If so, yes this does need to get added. Or we message to user the implications of renaming. I go with just changing reference for them.
> > @SPRCoreDump make sense? @neelneelneel inline with your thinking?
> 
> Was not aware of the reference issues due to this all being novel to me. I tried for a while to find a way to directly change the ID of the query directly, but I couldn't find an ActionMessage to pass to state.dispatch{} that would allow me to do such a thing...
> 
> Can I get an example of what would go wrong with a duplicate + delete? Also, is this an issue when duplicating as well?

You could do this multiple ways

Thought process 1:
- [ ] First Add a New Query with the new identifier 
- [ ] Replace the cells in this new query with the cells from the query we are going to delete
- [ ] Delete the old query

Thought process 2: (May be easier)
- [x] Create a new action that is called RENAME_QUERY (takes in a queryId, and a new name)
- This will first create a new query, but validate if the name is a uniq identifier.  Then take the contents of the old query and replace that in my newly made query.


Extra steps that are needed:
Replace all references of the old query id to the new id (this is where i think a new action would be better suited thought process 2)

Attached is a vid that outlines it, also a blocks.json that i explain what to do with in the video

[old-vars (1).json](https://github.com/user-attachments/files/16382169/old-vars.1.json)

https://deloitte.zoom.us/rec/share/LDzWsYtS6_2aowAsWg9M5UjTih08r-kJ-I37dI63FCOZ20EHSaRbA4t2ESZRqXoo.7A6h0SqFlamQ-4M-?startTime=1721930183000
Passcode: Y3O@ZLzr

neelneelneel:
Resolves #212

neelneelneel (changes requested):
Looks good for the most part. Can you

- [ ] Allow a `title` prop to be passed in to NewQueryOverlay.tsx that can change the title 'New Query' -> 'Rename Query'
- [ ] Show a warning that reference values won't be updated (Similar to Edit Variable)

PR #216, opened 2024-07-24
210 restrict extension type

### LINKED ISSUE #210: [TASK] Restrict extension type setting on the Upload block
Provide a new setting on the upload block that allows you to restrict what extensions are uploaded
/upload-block/config.tsx
Setting options will point to a constant.

That will be a list of file types that are safe (Please research into this)

/upload-block/UploadBlock.tsx

Upload Block will pull in new setting to tie to **accepts **property on the

rajivmaliDC:
created SelectSettings component to handle extensions
Updated config.tsx to add extension input with SelectSettings component 
updated UploadBlock.tsx to validate file extensions

To test the feature:

Create a drag and drop app 
add upload block
Select the upload block 
go to extension field
select multiple extensions
click save.

go back to the app
click on choose file 
file type should be auto filtered by custom files(with list of extension user specified)
user should be able to upload file with extension listed on custom file

 click on custom file and change it to "All Files(*.*)"
select the file type that is not listed on Custom file
user should not be able to upload it


johbaxter:
Will review by Friday

packages/client/src/components/block-defaults/upload-block/UploadBlock.tsx
```diff
@@ -20,12 +20,14 @@ export interface UploadBlockDef extends BlockDef<'upload'> {
...
         disabled: boolean;
         hint?: string;
+        extensions?: string[];
     };
 }
 
 export const UploadBlock: BlockComponent = observer(({ id }) => {
     const { attrs, data, setData, uploadFile, listeners } =
         useBlock<UploadBlockDef>(id);
+    const [acceptExtensions] = useState(data.extensions);
```
johbaxter: Do you need to useState?

packages/client/src/components/block-defaults/upload-block/UploadBlock.tsx
```diff
@@ -59,6 +61,18 @@ export const UploadBlock: BlockComponent = observer(({ id }) => {
...
             // save it as the value
             setData('value', fileLocation);
+
+            // extract te file extension from the file name
+            const fileExtension = `.${file.name
+                .split('.')
+                .pop()
+                .toLowerCase()}`;
+
+            // clear the value if fileExtension does not exist
```
johbaxter: Can we do this extension check before we upload the file on ln:49

packages/client/src/components/block-defaults/upload-block/config.tsx
```diff
@@ -45,6 +74,17 @@ export const config: BlockConfig<UploadBlockDef> = {
...
                     ),
                 },
+                {
+                    description: 'Extensions',
+                    render: ({ id }) => (
+                        <SelectSettings
+                            id={id}
+                            label="Extensions"
+                            path={'extensions'}
+                            extensions={FileTypes}
```
johbaxter: Can we rename this to options? Needs to be genericized as a component we can use to pass other options that we have stored as constants, will not always be extensions.

We also may have a Select Block Settings component already, if that is the case please use that one over this.  It would be in `/client/src/components/block-settings

^ Take a look at SelectInputSettings.tsx can we use this?

packages/client/src/components/block-settings/shared/SelectSettings.tsx
```diff
@@ -0,0 +1,118 @@
+import { useEffect, useMemo, useRef, useState } from 'react';
```
johbaxter, line 1: Like comment states above if we have a component that already does this, please use that one instead.

johbaxter (changes requested):

packages/client/src/components/block-defaults/upload-block/config.tsx
rajivmaliDC: updated extensions to Option to make it more generic for reusability

SelectInputSettings is designed for single option selection

packages/client/src/components/block-defaults/upload-block/UploadBlock.tsx
rajivmaliDC: moved up the extension check

packages/client/src/components/block-defaults/upload-block/UploadBlock.tsx
rajivmaliDC: removed useState

johbaxter:
image


@rajivmaliDC Can you ensure the Upload Setting takes those extensions into account as well

rajivmaliDC:
added inputProps to uploadSettings.tsx

PR #220, opened 2024-07-25
[BUG] Login page is not dynamic to the values in social.properties

### LINKED ISSUE #209: [TASK] Show/hide values on the logins based on /config
### Description 

The login page is not dynamic to the social.properties values. If you change the <auth_type>_login values from true/false it doesn't affect/change the UI

### Existing

- LoginPage.tsx
- config.store.ts

### Reference

https://github.com/SEMOSS/community/issues/11

### Tasks

- [x] Utilize `loginsAllow` from `/config` to dynamically set the login options

kgorbell:


neelneelneel:
Resolves #209

packages/client/src/pages/LoginPage.tsx
```diff
@@ -279,8 +281,28 @@ export const LoginPage = observer(() => {
...
     });
+    useEffect(() => {
+        fetchLoginTypesAllowed();
+    }, []);
 
-    const location = useLocation();
+    /**
+     * Find which types of logins are available to the user.
+     */
+    const fetchLoginTypesAllowed = async () => {
```
neelneelneel: `configStore` should already have this information. Another network call should not be necessary

packages/client/src/pages/LoginPage.tsx
```diff
@@ -238,10 +238,12 @@ interface TypeUserRegister {
  * LoginPage
  */
 export const LoginPage = observer(() => {
-    const { configStore } = useRootStore();
+    const { configStore, monolithStore } = useRootStore();
+    const location = useLocation();
 
     const [forgotPassword, setForgotPassword] = useState(false);
-    const [loginType, setLoginType] = useState('Native');
+    const [availableLogins, setAvailableLogins] = useState<string[]>([]);
```
neelneelneel: You could just leave it in the `loginsAllowed` object and check if they key exists

neelneelneel (changes requested):
There is also an issue where Microsoft is always visible . I believe it's related to the `providers` arrayreturned by the `configStore`.

neelneelneel (approved):

PR #252, opened 2024-08-13
246 filters in url params

### LINKED ISSUE #246: [TASK] Allow filters on the catalog pages to be url addressable
### Description 

Allow filters to be url addressable (as url parameters)

image

### Existing

`EngineCatalogPage.tsx`
`HomePage.tsx`

### Reference

This issue references. . .

### Tasks

- [ ] Allow for single and multiple selection of filters
- [ ] Fix engine catalog on `EngineCatalogPage.tsx`
- [ ] Fix home page on `HomePage.tsx`

kgorbell:
Please note that while working through this ticket, I came across a pre-existing bug related to the FilterBox component on the EngineCatalogPage that I was unable to address in this ticket due to time constraints.

**The bug:** when a user tabs through the various Catalog pages (I.E., models, engines., etc..), the state of the filters is maintained regardless of if that filter applies to the page or not. Additionally, let's say a user selects a filter on the Models page, then navigates to another catalog page and back to the Models page; the previous filters state in the FilterBox is maintained, but the filter is not applied to the fetch for the Models listed on the page. This is because the `EngineCatalogPage.tsx` is not aware of the state being maintained in the `FilterBox.tsx`.
**Possible Solutions:** either the FilterBox's state needs to be reset if the Catalog Type changes, OR, the FilterBox needs to relay its state back to the parent when the Catalog Type changes. 
Between the two, I feel that clearing the filters on any page change would be the better UX.

### DIFF
--- a/packages/client/src/components/ui/Filterbox/Filterbox.tsx
+++ b/packages/client/src/components/ui/Filterbox/Filterbox.tsx
@@ -287,7 +287,6 @@ export const Filterbox = (props: FilterboxProps) => {
             }
         });
         // Pass filters to parent
-        console.log('HANDLING IT', constructedFilters);
         onChange(constructedFilters);
         // Update query params in the URL
         setSearchParams(constructedFilters);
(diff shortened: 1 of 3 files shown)


johbaxter:
image
image

Good stuff ty @kgorbell

PR #286, opened 2024-09-18
Feature/multi select dropdown enhancement

22mayank:


packages/client/src/components/block-defaults/select-block/SelectBlock.tsx
```diff
@@ -16,9 +16,10 @@ import { optimize } from 'webpack';
 export interface SelectBlockDef extends BlockDef<'select'> {
     widget: 'select';
     data: {
+        isMulti: boolean;
         style: CSSProperties;
         label: string;
-        value: string;
+        value: any;
```
anurag91jain: To check if string | string [] is possible instead of any

anurag91jain (approved):

PR #292, opened 2024-09-23
feat(client): add json viewer to notebook cell output, operations and variables

### LINKED ISSUE #263: [TASK] Implement JSON Viewer to show and copy values
### Description 

We need a better UI to view properties on the JSON tied to our notebook queries and variables as well as our blocks

### Existing

https://viewer.textea.io/

### Reference

This issue references. . .

### Tasks

- [ ] Add package from above, for our JSON variables on our VariablePreview component
- [ ] Add package from above, for Notebook cells if the Output is JSON

JoshRome26:
Added in the json viewer to the preview for the add variable modal and for outputs in notebook cells if valid json outputs.


packages/client/src/components/notebook/VariablePreview.tsx
```diff
@@ -151,12 +161,16 @@ export const VariablePreview = observer((props: VariablePreviewProps) => {
...
+                {val !== null ? (
+                    <Stack direction="row">
+                        <Typography variant="body2" fontWeight="bold">
+                            Value:{' '}
+                        </Typography>
+                        <Typography variant="body2">{val}</Typography>
+                    </Stack>
+                ) : (
+                    <></>
+                )}
```
trevorzylks, line 171: Can you truncate this with the following? At first pass, I don't believe you need to return a fragment.
```
    {val !== null && (
                    <Stack direction="row">
                        <Typography variant="body2" fontWeight="bold">
                            Value:{' '}
                        </Typography>
                        <Typography variant="body2">{val}</Typography>
                    </Stack>
                )}
```

packages/client/src/components/notebook/NotebookCellConsole.tsx
```diff
@@ -15,14 +16,35 @@ interface ConsoleProps {
 
 export const NotebookCellConsole = (props: ConsoleProps) => {
     const { messages } = props;
+    const isOutputJSON = (str: string) => {
```
trevorzylks: Is this a utility function that can be used elsewhere in the application? If so, move it into a util folder -- otherwise disregard.

packages/client/src/components/notebook/AddVariablePopover.tsx
```diff
@@ -150,6 +152,41 @@ export const AddVariablePopover = observer((props: AddVariablePopoverProps) => {
 
     let alreadyAliased = false;
 
+    // TODO: Potentially Look into adding json viewer library for removing Editor.
```
trevorzylks: Do we want to keep these instances of commented out code?

package.json
```diff
@@ -23,6 +23,10 @@
         "npm-run-all": "^4.1.5"
     },
     "dependencies": {
+        "@emotion/react": "^11.13.3",
```
trevorzylks: Do these need to be installed at the top level, as opposed to within the client? And, if not, should the versions of these packages match what is in the client's package.json? Why two separate versions?

packages/client/src/components/notebook/VariablePreview.tsx
JoshRome26: Updated to not return a fragment.

packages/client/src/components/notebook/NotebookCellConsole.tsx
JoshRome26: Moved to utility folder.

packages/client/src/components/notebook/AddVariablePopover.tsx
JoshRome26: Removed commented out code, left TODO.

package.json
JoshRome26: Moved json-viewer into client package, removed instances in top level package.

packages/client/src/components/notebook/AddVariablePopover.tsx
```diff
@@ -475,8 +479,19 @@ export const AddVariablePopover = observer((props: AddVariablePopoverProps) => {
                             </Alert>
                         );
                     }
-                } else if (inputVariableTypeList.indexOf(variableType) > -1) {
-                    return JSON.stringify(variableInputValue);
+                } else if (
+                    variableType === 'JSON' ||
```
antgibson96: If we edit a variable, seems like it isn't respecting this if statement

antgibson96: Might be a good place to put your utility function

packages/client/src/components/notebook/VariablePreview.tsx
```diff
@@ -116,12 +117,21 @@ export const VariablePreview = observer((props: VariablePreviewProps) => {
             }
         } else {
             const found = state.parseVariable(`{{${id}}}`);
-
-            return (
-                <Typography variant="body2" fontWeight="bold">
-                    {typeof found !== 'string' ? JSON.stringify(found) : found}{' '}
-                </Typography>
-            );
+            if (typeof found === 'string') {
```
antgibson96: Add check to see if string is a json string

packages/client/src/components/notebook/operations/DefaultOperation.tsx
```diff
@@ -24,6 +25,15 @@ export const DefaultOperation = observer(
             return <StyledJson>{output}</StyledJson>;
         }
 
+        if (typeof output === 'object') {
```
antgibson96: Should be able to reference your utility method here too

packages/client/src/components/notebook/operations/SuccessOperation.tsx
```diff
@@ -13,6 +14,15 @@ export const SuccessOperation = observer(
     (props: SuccessOperationProps): JSX.Element => {
         const { output } = props;
 
+        if (typeof output === 'object') {
```
antgibson96: Utility method here

antgibson96 (changes requested):
Looks like there are a few edge cases where your type comparisons aren't converting to json. Look into why that is and solidify those, hopefully through that utility method you made

packages/client/src/components/notebook/AddVariablePopover.tsx
JoshRome26: Confirmed that the statement works and added in utility function.

packages/client/src/components/notebook/operations/DefaultOperation.tsx
JoshRome26: Utility method was added to check if the json string is json, but can't be used here. Adding the utility here / casting the output to work with the function returns the output as a string; Which is not what we want.

packages/client/src/components/notebook/operations/SuccessOperation.tsx
JoshRome26: Utility method was added.

packages/client/src/components/notebook/operations/SuccessOperation.tsx
```diff
@@ -13,6 +15,15 @@ export const SuccessOperation = observer(
     (props: SuccessOperationProps): JSX.Element => {
         const { output } = props;
 
+        if (isOutputJSON(output)) {
```
antgibson96: make else statement

packages/client/src/components/notebook/AddVariablePopover.tsx
antgibson96: think we can also clean up these conditions

packages/client/src/components/notebook/NotebookCellConsole.tsx
```diff
@@ -15,9 +16,37 @@ interface ConsoleProps {
...
+                            value={JSON.parse(m)}
+                            displayDataTypes={true}
+                            displaySize={true}
+                            displayComma={true}
+                            rootName={false}
+                        />
+                    );
+                }
+                // Check used to convert "{'test': 'sample'}" string into valid JSONs
+                const stringCheck = m.replace(/'/g, '"');
```
antgibson96: move this to the utility function

packages/client/src/components/notebook/VariablePreview.tsx
```diff
@@ -151,12 +161,14 @@ export const VariablePreview = observer((props: VariablePreviewProps) => {
                     </Typography>
                     <Typography variant="body2">{variable.type}</Typography>
                 </Stack>
-                <Stack direction="row">
-                    <Typography variant="body2" fontWeight="bold">
-                        Value:{' '}
-                    </Typography>
-                    <Typography variant="body2">{val}</Typography>
-                </Stack>
+                {val !== null && (
```
antgibson96, line 164: what prompted this check?

packages/client/src/utility/general.ts
```diff
@@ -163,3 +163,15 @@ SECRET_KEY="${secretKey ? secretKey : '<your secret key>'}"
 `;
     }
 };
+
+/**
+ * @desc Checks if output and verify if its a JSON object
+ */
+export const isOutputJSON = (output: string) => {
```
antgibson96: make output less specific, more of a "give me anything and I'll tell you if it's json viewer compatible"

packages/client/src/components/notebook/VariablePreview.tsx
JoshRome26: This check is specifically used to remove showing the preview variable value twice, before it would show the editor view of the variable and then then a typography of the variable when you hovered over. 

Now it will only show the json-viewer if the value is an object. Which is what was requested for all objects.

packages/client/src/components/notebook/operations/SuccessOperation.tsx
JoshRome26: Added else.

packages/client/src/components/notebook/VariablePreview.tsx
JoshRome26: Added in utility check here.

packages/client/src/components/notebook/AddVariablePopover.tsx
JoshRome26: Added back original condition, now works with all variable types including json / array.

packages/client/src/components/notebook/operations/DefaultOperation.tsx
JoshRome26: Resolved this, we check if output is both string and object and then use utility function to resolve.

packages/client/src/utility/general.ts
JoshRome26: Refactored function to return object if its compatible and null if it isn't.

antgibson96 (approved):

johbaxter:
Merged in changes manually while on dev due to build errors to this branch.  Ty @JoshRome26 and @antgibson96

PR #303, opened 2024-09-27
284 task admin editadd user bug fixes

### LINKED ISSUE #284: [TASK] Admin Edit/Add User bug Fixes
### Description 

When adding and editing new users from admin member setting the following issues need to be fixed

- Form Validation reason needs to in red font 
- check if phone numbers are required 
- when editing the email, editUser required the parameter passed in to be newEmail not email. 

### Existing

### Reference

This issue references. . .
Screenshot 2024-09-16 at 2 10 24 PM

### Tasks

What are more specific tasks that a developer needs to take to consider this issue resolved.

mattfreshwaters:
For issue https://github.com/SEMOSS/semoss-ui/issues/284

Regex is fixed so form doesn't think XXX-XXX- is a valid phone number, accepst (XXX) XXX-XXXX or XXX-XXX-XXXX

Now sending 'newEmail' instead of 'email' to the backend for editUser

Error messages on the editUser form are now in red


johbaxter:
image

Is there a reason we don't pull in email into the modal. @mattfreshwaters

johbaxter:
image
image

@mattfreshwaters also the way we send it versus the way we receive it are totally different can we have some consistency

mattfreshwaters:
> image image
> 
> @mattfreshwaters also the way we send it versus the way we receive it are totally different can we have some consistency
@johbaxter 
Assuming you're referencing the phone number - yep can do that. Do we have a preference of what format? Figured for user sake it is easiest if they have different options to type in, but we can adjust and send a single format to the backend (i.e. 123-456-7890) for consistency - thoughts?.

mattfreshwaters:
@johbaxter backend PR here: https://github.com/SEMOSS/Semoss/pull/171
created an issue for it in Semoss as well - https://github.com/SEMOSS/Semoss/issues/140

packages/client/src/components/settings/UserAddOverlay.tsx
```diff
@@ -20,6 +20,10 @@ import { useForm, useFormState, Controller } from 'react-hook-form';
 import { useRootStore, useSettings } from '@/hooks';
 import { AxiosResponse } from 'axios';
 
+const StyledTypography = styled(Typography)({
+    color: 'red',
```
neelneelneel: Update to the 'error' color

packages/client/src/components/settings/UserAddOverlay.tsx
```diff
@@ -187,6 +199,12 @@ export const UserAddOverlay = (props: UserAddOverlayProps) => {
      */
     const editUser = handleSubmit(
         async (data: EditUserForm) => {
+            if (email !== user?.email) {
+                console.log('DING DING DING');
```
neelneelneel: Remove extra console.logs

neelneelneel (changes requested):

packages/client/src/components/settings/UserAddOverlay.tsx
mattfreshwaters: Just changed in newest commit

packages/client/src/components/settings/UserAddOverlay.tsx
mattfreshwaters: Should be gone now

neelneelneel (approved):

PR #305, opened 2024-09-30
feat(client): added custom css settings to page and container blocks

soumya-pratik:
image
image

Added custom css editor which will be part stateStore and can be appended to a style tag under head of document


neelneelneel:
Add the `rootStyle` to the page object instead of the store. This will allow the tool to have different style sheets per page when we enable multipage apps.

neelneelneel (changes requested):
Add the rootStyle to the page object instead of the store. This will allow the tool to have different style sheets per page when we enable multipage apps.

Issue #68, opened 2024-07-16
[TASK] UI Builder: Default Padding Updates and Bring Back Section


iclyn-taero:
Decided with Johnnathan, Kyle and Sarah:

- Updated default Page padding = 0px and gap = 0px to accomodate for grouped components like navs, footers and heros which should span entire page. 
- Update 'welcome block container" to be called "Main Container". This main container should always be present when a user initially loads UI builder. Height should default to 90%.
- Bring back Section/Grid as layout option

PR #319, opened 2024-10-09
feat(client): added sub containers under container as settings

soumya-pratik:


johbaxter:
@soumya-pratik what is the purpose of this, why wouldn't we just ask users to drag in containers within containers.

I think this just adds another complexity that's not needed.

ContainerBlock and SubContainerBlock seem to be the same code other than the added property for subcontainer on the container block

I personally don't think this is needed

anurag91jain:
1. The main reason was to make it intuitive. As of now, an editor has to drop a lot of containers within containers to manage the look and feel of the app which makes it tedious. For example, for creating a simple application with sidebar and a chatbox at the end, we had to drag and drop 5 containers. 
2. For now it is not a strict rule that containers will ALWAYS have a subcontainer.
3. We can also look at reducing some of the properties of the container as well, such as dimensions and border. 
4. If someone wants to create a card type layout, having sub containers would make it simpler.
Salesforce implements something similar with lighting layout. (https://developer.salesforce.com/docs/component-library/bundle/lightning-layout)
@johbaxter

soumya-pratik:
@johbaxter I have fixed the bug you mentioned in the recording. You can check it.

neelneelneel:
@johbaxter Is https://github.com/orgs/SEMOSS/projects/4/views/2?pane=issue&itemId=71153233&issue=SEMOSS%7Csemoss-ui%7C68 related?

johbaxter:
> @johbaxter Is #68 related?

Yup this is a mixture of this and the Grid, along with the compound blocks.

What was added to this, was a property on the container-block.  Where you update the count on a specific setting, and it will dynamically remove and adds new containers with state.dispatch.

neelneelneel:
Dependent on https://github.com/SEMOSS/Semoss/issues/254

johbaxter:
Not a viable approach

PR #321, opened 2024-10-10
fix(general.ts): resolve duplication issue

dlee32ee:
Fixed module/base duplication issue from the my profile page.


packages/client/src/utility/general.ts
```diff
@@ -148,14 +148,14 @@ server_connection=ai_server.RESTServer(
     secret_key="${
         secretKey ? secretKey : '<your access key>'
     }",             # example: "c2b3fae8-20d1-458c-8565-30ae935c4dfb"
-    base="${Env.ORIGIN}${Env.PATH}${Env.MODULE}/api"
+    base="${Env.ORIGIN}${Env.PATH}${process.env.MODULE}/api"
```
antgibson96, line 151: this should be the same value if we are setting them correctly, what exactly is the issue here?

PR #322, opened 2024-10-14
feat(client): add a select component for predefined engine types by Mayank Agarwal

22mayank:


packages/client/src/components/block-defaults/select-engine-block/config.tsx
```diff
@@ -0,0 +1,118 @@
...
+        label: 'Example Select Engine Input',
+        hint: '',
+        options: [],
+        required: false,
+        disabled: false,
+        loading: false,
+        optionLabel: '',
+        optionSublabel: '',
+        optionValue: '',
+        multiple: false,
```
neelneelneel: Remove extra / unnecessary attributes

packages/client/src/components/block-defaults/select-engine-block/config.tsx
```diff
@@ -0,0 +1,118 @@
...
+    },
+    listeners: {
+        onChange: [],
+    },
+    slots: {
+        content: [],
+    },
+    render: SelectEngineBlock,
+    icon: ViewList,
+    contentMenu: [
```
neelneelneel, line 41: Remove unnecessary settings

packages/client/src/components/block-defaults/select-engine-block/SelectEngineBlock.tsx
```diff
@@ -0,0 +1,167 @@
...
+    LinearProgress,
+    Stack,
+    TextField,
+    Typography,
+} from '@mui/material';
+import { optimize } from 'webpack';
+
+export interface SelectEngineBlockDef extends BlockDef<'select-engine'> {
+    widget: 'select-engine';
+    data: {
```
neelneelneel, line 17: Add a filter for the various catalog types.

neelneelneel (changes requested):
Remvo

22mayank:
Hi Neel, I have addressed the PR feedback for removing unused attributes & unnecessary settings. The filter for various catalog is already present. ( Attached screenshot )
image

neelneelneel (changes requested):
Few changes
- [ ] Move the call to get engines to `SelectEngineBlock.tsx`. This will enable infinite-loading
- [ ] Options can be removed, so Data should be defined as
```
export interface SelectEngineBlockDef extends BlockDef<'select-engine'> {
    widget: 'select-engine';
    data: {
        style: CSSProperties;
        label: string;
        type: string // corresponds to engine type
        value: {name:string; id:string} // corresponds to name + id
        required: boolean;
        disabled: boolean;
        hint?: string;
        loading?: boolean;
    };
}
```

neelneelneel:
@themaherkhalil  - how should this interact with dependencies?

neelneelneel:
Per discussion, doesn't interact with dependencies as it is user-dependent

johbaxter:
Not a viable approach

PR #339, opened 2024-10-25
Audio Input UI block

soumya-pratik:
image

You can find the implementation recording here - https://amedeloitte.sharepoint.com/:v:/r/sites/AIServer/Shared%20Documents/General/Documents/CFG%20Platform%20Enhancement%20Implementation%20videos/[Audio-input-demo.mp4](https://amedeloitte.sharepoint.com/:v:/r/sites/AIServer/Shared%20Documents/General/Documents/CFG%20Platform%20Enhancement%20Implementation%20videos/Audio-input-demo.mp4?csf=1&web=1&e=n9pMkm)?csf=1&web=1&e=n9pMkm


johbaxter:
image

Hey @soumya-pratik we should have a setting on the right to interact with the element while in edit mode.

Looking at the code when you click the recording, and stop.  The transcript binds to the value setting. 

^This is good.  But can we have a spot on the settings menu to update the value property similar to how we do settings for other blocks

The setting on the right that you add should also have the voice recognition capabilities

packages/client/src/components/block-defaults/audio-input-block/config.tsx
```diff
@@ -0,0 +1,117 @@
...
+                            id={id}
+                            label="Loading"
+                            path="loading"
+                            queryPath="isLoading"
+                        />
+                    ),
+                },
+            ],
+        },
+    ],
```
johbaxter, line 62: We need a setting to update the "value" property.  This should be an input with a mic icon for the start adornment.

image

^along the lines of this

soumya-pratik:
> image
> 
> Hey @soumya-pratik we should have a setting on the right to interact with the element while in edit mode.
> 
> Looking at the code when you click the recording, and stop. The transcript binds to the value setting.
> 
> ^This is good. But can we have a spot on the settings menu to update the value property similar to how we do settings for other blocks
> 
> The setting on the right that you add should also have the voice recognition capabilities

@johbaxter I have updated the settings as per your comment. Please check once.
image

johbaxter:
Alright so your pretty much there.  But this setting component needs to have the same capability as the UI element itself.

image

The textfield that you show for the setting component needs to have the audio piece to it as well

soumya-pratik:
> Alright so your pretty much there. But this setting component needs to have the same capability as the UI element itself.
> 
> image
> 
> The textfield that you show for the setting component needs to have the audio piece to it as well

I have added the audio piece and it works same as the component. 
image

johbaxter (approved):
Looks good ty

PR #342, opened 2024-10-28
feat(client): add a radio group component

22mayank:


packages/client/src/components/block-defaults/radio-block/config.tsx
```diff
@@ -0,0 +1,347 @@
...
+                                    ...newOptions,
+                                ]);
+                                setNumAdditionalOptions('');
+                                setOptionLabels('');
+                            }
+                        };
+
+                        // Find the current option object for the selected value
+                        return (
+                            <Stack spacing={2}>
```
johbaxter: image

We need to rethink this.  Avoid the extra step, of setting a number of options.

Why don't we just have a singular field. That allows you to add options dynamically without having to specify the amount of options. Something like this.

image

The plus button would obviously add another option

packages/client/src/components/designer/menu/default-menu.ts
```diff
@@ -98,6 +99,34 @@ export const DEFAULT_MENU: DesignerMenuItem[] = [
...
+        json: {
+            widget: 'radio',
+            data: {
+                style: {
+                    padding: '4px',
+                },
+                value: '',
+                label: 'Radio Input',
+                isGroup: false,
+                options: [],
```
johbaxter: Can we set a default option.

options: [
{label: ''default', value: 'no value'}
]

packages/client/src/components/designer/menu/default-menu.ts
```diff
@@ -98,6 +99,34 @@ export const DEFAULT_MENU: DesignerMenuItem[] = [
...
+                size: 'medium',
+                direction: 'column',
+                color: 'primary',
+                labelPlacement: 'end',
+                required: false,
+                disabled: false,
+                optionLabel: '',
```
johbaxter: Lets get rid of optionLabel

packages/client/src/components/block-defaults/radio-block/config.tsx
```diff
@@ -0,0 +1,347 @@
...
+    data: {
+        style: {
+            padding: '4px',
+        },
+        value: 'yes',
+        label: 'Radio Group',
+        options: [
+            { label: 'Yes', value: 'yes' },
+            { label: 'No', value: 'no' },
+        ],
```
johbaxter, line 162: This interface for options is good, don't change this

johbaxter (changes requested):

tevanburen:
Could we give users the ability to edit (or at least delete) their options? As of right now, if I make a typo, I'd have to delete and recreate the entire component. One idea is to replace the top section with a kind-of table:

image

Other implementations would work as well

22mayank:
image


Added the Tabular format for options. Kept the Default option as disabled but can be deleted and if all the options are deleted the  Default option is recreated.

johbaxter:
> image
> 
> Added the Tabular format for options. Kept the Default option as disabled but can be deleted and if all the options are deleted the Default option is recreated.

Why did we get rid of the setting for selected value.

Also please, there should be no restrictions on setting the options.  

Think about how you would like to use it.

If you drag a radio group in, you want to edit the options (no limitations) whether it be deleting an option, adding a new one in, as well as resetting properties on those options

johbaxter:
@22mayank you got to test man.

1. I click radio group 
2. I add my own options
4. Then i click away from radio group
5. Then click back into radio group, all the options i created (step-2) disappear other than the first index

johbaxter (approved):

PR #343, opened 2024-10-30
Stepper Block

### LINKED ISSUE #324: [TASK] Stepper Block
### Description 

Add a block for material UI Stepper component and create settings to modify properties specific to Stepper
https://mui.com/material-ui/react-stepper/

### Existing

block-defaults/stepper-block
designer.constants

### Reference

N/A

### Tasks

- Modify StepperBlock.tsx to show a step for the list of steps specified on data.steps
- Modify StepperBlock.tsx onClick to set data.activeStep to the selected step
- Create Settings component to be able to set data.steps for new block
- Create settings for data.activeStep to allow you to set the current step through settings

j-arias-dev:
### Add a block for material UI Stepper component and create settings to modify properties specific to Stepper

image


packages/client/src/components/block-defaults/stepper-block/config.tsx
```diff
@@ -36,11 +28,21 @@ export const config: BlockConfig<StepperBlockDef> = {
...
+                    description: 'Active Step',
+                    render: ({ id }) => (
+                        <InputSettings
                             id={id}
-                            label="Enable Multi Select"
-                            path="multiple"
-                            description="This setting will enable the multi-select feature on the select input"
+                            label="Active Step"
+                            path="active"
+                            type="number"
```
johbaxter, line 45: image
image

I need to be able to reference a variable from the notebook to determine what step is active.  The way this is setup at the moment, does not allow any dynamic behavior.

Meaning a button click outside of the stepper cannot determine the nextStep.

The StepperBlock, should parseInt() if we are worried about the setting it to the right data type

For Example: 

image

packages/client/src/components/block-defaults/stepper-block/StepperBlock.tsx
```diff
@@ -1,31 +1,44 @@
...
+            <Stepper {...attrs} activeStep={active}>
+                {data.steps?.map((obj, i) => {
+                    return (
+                        <Step key={`${id}--${obj.name}--${i}`}>
+                            <StepLabel>{obj.name}</StepLabel>
+                        </Step>
+                    );
+                })}
+            </Stepper>
+            {Number(active) >= data.steps?.length ? (
```
johbaxter, line 33: This whole block of showing the content, should not be apart of this block. (Content will be rendered by another block, where we will use a conditional) based on the activeStep)

Lines: 33-41

johbaxter (changes requested):

PR #344, opened 2024-10-30
335 Error Message for Generating a key on a user's profile page

### LINKED ISSUE #335: [TASK] Update FE to display the BE error message when user tries to create a key without proper access
### Description 

Update the FE error message that displays to the user with the BE error message. The BE error message is more contextually helpful to the user. 

Screenshot: 
image (18)

### Existing

n/a

### Reference
 
n/a

### Tasks

n/a

kgorbell:
Important note for testing:
This fix had to be addressed at the root of app where the Axios interceptor handles processing the returns when an API call errors out; and therefore requires testing that is not exclusive to the My Profile Page.

### DIFF
--- a/packages/client/src/App.tsx
+++ b/packages/client/src/App.tsx
@@ -26,19 +26,17 @@ axios.interceptors.response.use(
             }
         }
 
-        // Exception for returning the errorMessage provided via the API if available.
         const apiMessage = error.response?.data?.errorMessage;
         if (apiMessage && typeof apiMessage === 'string') {
+            // Exception for returning the errorMessage provided via the API if available.
             return Promise.reject(apiMessage);
-        }
-
-        // return the message if it exists
-        if (error.message) {
+        } else if (error.message) {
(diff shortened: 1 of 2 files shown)


packages/client/src/App.tsx
```diff
@@ -27,6 +26,12 @@ axios.interceptors.response.use(
             }
         }
 
+        // Exception for returning the errorMessage provided via the API if available.
+        const apiMessage = error.response?.data?.errorMessage;
```
antgibson96, line 29: looks like the check above this on line 18 is similar to part of this, would it make more sense to add it into that if/else ladder?

kgorbell: If anything, I feel this would be better grouped in an if/else with the statements below (36 & 40) since they are also using logic to determine what message should be returned in the error. But, since the previous dev had the if statement separated with comments to explain the if/else, I followed that pattern in my code here.

PR #356, opened 2024-11-07
Feature/table block implementation v1

22mayank:


neelneelneel (changes requested):
This functionality should be merged with [TableBlock.tsx](https://github.com/SEMOSS/semoss-ui/blob/dev/packages/client/src/components/block-defaults/table-block/TableBlock.tsx) instead of being a new block. Please do not forget the [MigrationManager.ts](https://github.com/SEMOSS/semoss-ui/blob/dev/packages/client/src/stores/state/migration/MigrationManager.ts)

neelneelneel:
Covered by https://github.com/SEMOSS/semoss-ui/pull/391

PR #357, opened 2024-11-08
feat(client): add a modal component

22mayank:


johbaxter (changes requested):
Discussed course of action on call 12/5/24

- Remove Button that is used to open modal on the Block itself.  The open state of the modal should be determined by a notebook or event.

- Modal Actions should be dropped in as a block rather than it being apart of the block

- Keep design mode, in order to design modal without relying on the open state of the block

johbaxter (approved):

PR #376, opened 2024-11-13
Echart ai jumpstart barchart 

bannaarisamy-shanmugham-kanini:
Bar Chart Initial setup and this pull request contains:
**1.Bar Chart Features**

1. Legend - toggle on/off
2. Customize value labels
3. Display value labels 
4. Tooltips  - toggle on/off 
5. Trendlines - toggles on/off 

**2.Bar Chart Styling**

1. Change bar colors 
2. Add chart title
3. Set font size and color 
4. Edit bar width

**3.Bar Chart Axis**

1. Show/hide each axis
2. Add axis titles
3. Edit label font size
4. Apend/prepend value to labels 
5. Show tick marks 
6. Rotate axis values (by a degree amount)


packages/client/src/components/block-defaults/echart-visualization-block/EChartVisualizationBlock.tsx
```diff
@@ -0,0 +1,76 @@
...
+})<{ error?: boolean }>(({ error = false, theme }) => ({
+    height: '20vh',
+    width: '80vh',
+    color: error ? theme.palette.error.main : 'unset',
+}));
+
+export interface EChartVisualizationBlockDef {
+    widget: 'e-chart';
+    data: {
+        option: any;
```
neelneelneel: Could you type this out in typescript?

packages/client/src/components/block-defaults/echart-visualization-block/EChartVisualizationBlock.tsx
```diff
@@ -0,0 +1,76 @@
...
+    function echartsLoaded(echartInstance) {
+        console.log('chart loaded', echartInstance);
+        echartInstance.dispatchAction({
+            type: 'takeGlobalCursor',
+            key: 'dataZoomSelect',
+            // Activate or inactivate.
+            dataZoomSelectActive: true,
+        });
+        echartInstance.on('click', function () {
+            console.log('click detected');
```
neelneelneel: You'll need to add / trigger an action via the listeners

packages/client/src/components/block-defaults/echart-visualization-block/ChartAxis.tsx
```diff
@@ -0,0 +1,182 @@
...
+        }
+        setChartAxisState((prevState) => {
+            return {
+                ...prevState,
+                ...chartAxisStateData,
+            };
+        });
+    }, []);
+
+    function updateXAxisState(axisState, axisValue) {
```
neelneelneel: Can you comment all methods?

packages/client/src/components/block-defaults/echart-visualization-block/ChartStyling.tsx
```diff
@@ -0,0 +1,288 @@
...
+                    {chartAlignment.map((chart, index) => {
+                        return (
+                            <Select.Item value={chart.value}>
+                                {chart.label}
+                            </Select.Item>
+                        );
+                    })}
+                </StyledSelect>
+            </div>
+            <div
```
neelneelneel: Can you remove the inline styles? And shift to using the components?

neelneelneel (changes requested):

packages/client/src/components/block-defaults/echart-visualization-block/EChartVisualizationBlock.tsx
bannaarisamy-shanmugham-kanini: can you please suggest me with a sample piece of code ? so I can follow the process

packages/client/src/components/block-defaults/echart-visualization-block/EChartVisualizationBlock.tsx
bannaarisamy-shanmugham-kanini: I am resolving with empty object, as option will have different keys, and all may not be needed.

neelneelneel:
@ehynd - Does this pass your review?

neelneelneel (changes requested):
- [x] Hide Old Vega Charts
- [x] Remove the Echarts from the name (so it has a nice name)
- [x] Consolidate `echarts-visualization-block` and `echarts-visualization-block-bk`
- [x] Fix dragging a block from the left to the right. Currently has an error: 
image

packages/client/src/components/block-settings/shared/EchartSettings.tsx
```diff
@@ -0,0 +1,259 @@
...
+                        suggestions: generateSuggestions(replaceRange),
+                    };
+                },
+                triggerCharacters: ['"{'],
+            });
+        };
+
+        return (
+            <Suspense fallback={<>...</>}>
+                <Editor
```
neelneelneel: Not sure if this is needed anymore.

packages/client/src/stores/state/state.store.ts
```diff
@@ -513,6 +513,18 @@ export class StateStore {
...
+                if (stringValue !== undefined) {
+                    let jsonValue = JSON.parse(stringValue);
+                    if (jsonValue !== undefined)
+                        value = JSON.parse(stringValue);
+                }
+            } catch (e) {
+                console.log('error occurred while parsing');
+                return value;
+            }
+
```
neelneelneel: We shouldn't make modification to this file. What is the purpose of this?

neelneelneel (changes requested):

packages/client/src/stores/state/state.store.ts
bannaarisamy-shanmugham-kanini: While linking variables to chart, at Echarts entire json came up as a string. So added the above fix. Now i removed the above fix

bannaarisamy-shanmugham-kanini: While getting the variable value, it is coming as string. to apply the json string to echart, I parsed it and applied the value. After parsing, it worked for both vega and echarts. Now I am removing this changes completely.

packages/client/src/components/block-settings/shared/EchartSettings.tsx
bannaarisamy-shanmugham-kanini: I have removed this file.

ehynd (dismissed):
@bannaarisamy-shanmugham-kanini  

a few functionality details on bar chart: 

1. could you please add the ability to resize the chart? @JaganSomannaKanini this is going to need to be implemented for pie chart as well 

Ex: I put two charts inside a container but can't size them to fit inside the container, so they overlap with the pie chart: 
image

These are the present options in container resizing (found if you add a container block and then hit block settings) 
image


2. The tooltip hover stays showing the last value it hovered over. Ex: In this screenshot, I hovered over the fourth bar and saw the tool tip. But, when I moved off the bar to another part of the screen (not another bar), it stayed on the screen. I want to only see the tooltip when I am hovering on the bar 

image

3. is there anyway to make the y field multi-select so the same chart block could be used for a multi-bar chart? Ex of what I'm hoping for:  In legacy system, I can select two fields for Y so the value for each populates with each age. 
image

bannaarisamy-shanmugham-kanini:
@ehynd , I have resolved following items:
Tooltip: tooltip is being displayed when not being hovered on chart - fixed
Multiple Y Axis: Single selection to multiple selection of Y Axis fields is now fixed and it is working.

ehynd:
@bannaarisamy-shanmugham-kanini I'm seeing a bug with the multi-select.  I selected a field called Admin_Task for the x and then both Expired and Upcoming for Y. The tooltip seems to be reflecting the numbers I'm expecting, but it shows "Admin_Task" instead of Expired.  The Y-Axis label is also defaulting to "Expired".  That can be switched easily with the setting the axis title, but wondering if those two are getting switched somehow? 
image

Would you also be able to update the Set Bar Color to accommodate multi select? Ex: I'd like to be able to set the color of the bars for each Y Category (Upcoming and Expired).

ehynd:
@bannaarisamy-shanmugham-kanini 

I pulled the branch with the added UI design - but something's not quite right. Getting a bunch of errors, mostly type errors.  It's resulting in not being able to use the charts at all: 
image

can you please take a look at this Wednesday and try to troubleshoot?

ehynd:
1-21 update: 
-need to explore sizing container element, not child visualization element
-should explore data.style within ContainerBlock.tsx, buildDimensionsSection() in config.tsx and block-defaults.shared.tsx to expose menu

ehynd:
1-24 Update
Sizing working as expected for bar chart! 

One bar chart bug:
I have a dataset in my data notebook. After I run the code for it, I am using that frame to build the bar chart. Then, I save the app and close it. 
When I reopen the app and run the data notebook, the bar chart does not populate again, it stays blank. 
Could you please take a look at this? Expected behavior is when I reopen the app and run the saved data notebook, the chart will populate again as well. 

Pie Chart:
@bannaarisamy-shanmugham-kanini and @JaganSomannaKanini - could one of you make sure to continue working the pie chart as well? The chart stays blank when selecting data fields now.

ehynd:
1-28 Update: 
These two charts are so close, almost to the end!

@bannaarisamy-shanmugham-kanini on bar:

Priority 1: you please add a toggle switch in the Edit X-Axis and Edit Y-Axis for the "Zoom Axis" feature? It is currently defaulting to on for Y-Axis, but there is no way to turn it off. 
Priority 2: Could you please work on adding in the color palette tool, including the custom palette. Sayali designed this screen in the Figma.
Priority 3: Could you please add the legend tool in, creating another tile for it to match the other tools with a toggle switch for on/off?

@JaganSomannaKanini on pie:

Priority 1: could you please take another look at the saving and re rendering? Maybe check with Banari on this? Behavior I am seeing right now is I can create the pie chart, save the pie chart, reload the app, and get this error: 
image
Priority 2: Could you also work with Banari on the sizing so that the charts follow the same approach? 
Priority 3: Please work on adding in color palette to tools as well, again making the approach the same as how it gets implemented in bar chart. 
Priority 4: Please make sure the legend tool gets transferred over.

packages/client/src/components/block-defaults/page-block/PageBlock.tsx
neelneelneel, line 1: Why was this added in?

packages/client/src/components/designer/Screen.tsx
neelneelneel, line 1: Why was this changed?

packages/client/src/stores/state/state.store.ts
neelneelneel, line 1: We shouldn't update the store. Create a utility file and share this method across.

neelneelneel (dismissed):
Few Additional Bugs/Issues/Comments:
 - [x] Filter doesn't work (right click)
 - [ ] Try to reuse settings in `client/src/components/block-settings/*` when building menu components. These are used across the different blocks and helps with consistency.
 - [x] We don't always have to have an "execute" button
 - [ ] Add in echarts typings

Smaller Details:
 - [x] Remove Unused Files / Code (ie various methods/components in Bar.tsx)
 - [x] Unused unnecessary change (ie: ButtonBlock.tsx)
 - [x] Rename components to be Title Case (ie: pie-chart-feature.tsx)
 - [x] Remove extra console.logs (ie: useFrame.ts)
 - [x] Expand comments. All props must be commented and typed.

ehynd:
1-31 
@JaganSomannaKanini pie changes look great! 

could you please take one more look at sizing? @bannaarisamy-shanmugham-kanini could you assist on this with the method you got it to work in bar chart?
Behavior I'm seeing is while the blue outline around the chart changes size, but the actual pie chart itself doesn't change at all to size with it. To replicate if necessary, I placed a container and then two charts with it. In the screenshot, the teal outline is a container that the the charts are within.  In this, the chart on top has a much lower set height than the bottom chart. The actual pies are the same while only the boxes are reflecting the size difference. 

image 

otherwise these look great and just need the color palette. thank you both!!

neelneelneel:
@bannaarisamy-shanmugham-kanini - Could you please address some of the open items on the PR? Still seeing a few errors when opening the files

bannaarisamy-shanmugham-kanini:
Hi @neelneelneel and @ehynd ,

In CustomizeValueLabels,tsx, I faced following issue when I tried to utilize ColorSettings.tsx component for updating color to a specific series in json instead of just a customized textfield,
in the following image, you could see a customized textfield, to update the color for a series:
image

I would like to use something like `<ColorSettings id={id} label='Series1 Color' path='series.0.label.color' />` instead of custom textfield for the following path: series[0]['label']['color'].
but the path value **series.0.label.color** is showing undefined in the useMemo function while creating the computedValue.
Please give some suggestions on this.

ehynd:
^^ @johbaxter could you take a look at this q?

bannaarisamy-shanmugham-kanini:
> Hi @neelneelneel and @ehynd ,
> 
> In CustomizeValueLabels,tsx, I faced following issue when I tried to utilize ColorSettings.tsx component for updating color to a specific series in json instead of just a customized textfield, in the following image, you could see a customized textfield, to update the color for a series: image
> 
> I would like to use something like `<ColorSettings id={id} label='Series1 Color' path='series.0.label.color' />` instead of custom textfield for the following path: series[0]['label']['color']. but the path value **series.0.label.color** is showing undefined in the useMemo function while creating the computedValue. Please give some suggestions on this.

Hi @ehynd , @neelneelneel ,
I am able to figure out a solution for setting path for series.0.label.color scenario. I have committed my changes related to colorpicker in color-palatte-design-dev branch. I will work on color palatte changes and then I will finally commit every changes for the color picker and color palette.

johbaxter (approved):

PR #386, opened 2024-11-15
feat(client): extend audio input block component

22mayank:


johbaxter (changes requested):
How does this conflict with the addition to this component that @soumya-pratik made

johbaxter:
image

Going with Option #2 will create new fresh PR

Issue #350, opened 2024-11-05
DRIVER property is missing in PGvector Database smss file


jayesh00949:
### Description 

For enabling "use connection pooling" option in pgvector DB, "DRIVER" property is also needed from smss file which is not getting added in smss file when we are creating a connection.

### Steps to reproduce
1. Open cfgai platfrom
2. Click on open vector.
3. Click on any existing PGvector db or create a new one.
4. After opening the db click on smss.
5. Observe the missing driver property

### Reference

![Screenshot (12)](https://github.com/user-attachments/assets/f6c620b1-9c65-4e28-9f12-8478b2a11ba4)

kunal0137:
Need to pass this during the creation of a vector db
DRIVER  org.postgresql.Driver

susanta96:
Closing the issue, as this got fixed.

PR #394, opened 2024-11-21
fix(client): added driver property in pgvector

susanta96:
DRIVER Property required for the use_connection_pooling. In PGVector the driver property was not passed from the frontend. Hope this solves the issue.


susanta96:
Closes #350

themaherkhalil:
This is an issue with the BE. We should be setting the driver properly from the RDBMS_TYPE field as that is our standard enum. Resolved in https://github.com/SEMOSS/Semoss/pull/273. Will close out this ticket.

PR #412, opened 2024-12-06
Allow node more space in packages/ui scripts

tevanburen:
pnpm run build was giving 
`FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory`
for some users. 

Increasing the node space seems to have fixed it.

### DIFF
--- a/packages/ui/package.json
+++ b/packages/ui/package.json
@@ -12,8 +12,8 @@
     "scripts": {
         "start": "pnpm run storybook",
         "dev": "run-p build:dev storybook",
-        "build": "rollup -c",
-        "build:dev": "rollup -w -c",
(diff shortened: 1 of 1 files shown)


susanta96:
Hi @johbaxter , Ui folder does take lots of time to build, there are few things I am trying to improve on that. If you want, I can create a PR (need some time to test it properly).

tevanburen:
I second that it takes much longer to build than would be ideal; it's especially noticeable on low-quality machines. The development experience would be much better if the runtime were improved

susanta96:
Working on it. I need to test it properly .

johbaxter:
Resolved in https://github.com/SEMOSS/semoss-ui/pull/417

PR #423, opened 2024-12-13
feat(client): added routing updates and multi page view

### LINKED ISSUE #408: [TASK] Introduce routing for Drag and Drop Apps
### Description 

As apps need to not be limited by single page view, we will need to introduce multi-page flexibility

Use react-router-dom, **do not add new routing library**

**SIDE NOTE**

As for directing user from one route to another it should only happen while users are actually using the app, versus making direct edits to the app.

Reason being if we reroute while making edits we will not be able to save the changes that were made to the page previous to the rerouting

_Edit Mode_
http://localhost:9090/semoss-ui/packages/client/dist/#/workspace/ae394abf-48c5-4de9-abdd-f03a2caf1a7e

versus

_Use App Mode_
http://localhost:9090/semoss-ui/packages/client/dist/#/app/ae394abf-48c5-4de9-abdd-f03a2caf1a7e
http://localhost:9090/semoss-ui/packages/client/dist/#/s/ae394abf-48c5-4de9-abdd-f03a2caf1a7e

### Existing

[BlocksRenderer.tsx](https://github.com/SEMOSS/semoss-ui/blob/dev/packages/client/src/components/blocks-workspace/BlocksRenderer.tsx)
[BlocksWorkspace.tsx](https://github.com/SEMOSS/semoss-ui/blob/dev/packages/client/src/components/blocks-workspace/BlocksWorkspace.tsx)
[NotebookExplorerPanel.tsx](https://github.com/SEMOSS/semoss-ui/blob/dev/packages/client/src/components/blocks-workspace/panels/NotebookExplorerPanel.tsx)

### Tasks

- [ ] Add dynamic routes to the BlocksRenderer `*`
- [ ] Add ability to add and view all pages via the layers panel while in edit mode `**`
- [ ] Modify DesignerViewerPanel to be able to show panels for multiple pages `***`

`*` Once the state store is initialized, you should have access to all page blocks that have been appended to state tied to the app.  Create routes for each of the page blocks, each of those blocks should have a data property to be able to set the path for the `<Route path={pageJson.data.route}/>`  We will need to add a setting on all page blocks to specify the route we want for that specific page.

`**` While in edit mode, add the ability to add pages to the blocks.json via a button click "Add Page", this button should first append a new page to the blocks.json as well as mimic similar behavior to the `NotebookExplorerPanel.tsx` within the `createPanel` method that utilizes `model.doAction()` and `Actions.addNode()`. the most important part is you need to specify what component _(task_3 from above)_ to render and the new page id within the config so the Factory can be aware of what page to render.

`***` In `BlocksWorkspace.tsx` we will modify the condition in the FACTORY for designer, the <DesignerPanel /> component will take a new property for page_id, which should come from `config.id` similarily to NotebookViewerPanel.  In the Designer component within it we will drill that same property down to then pass that page_id to the Renderer which currently renders based off of `designer.rendered` (Refactor if needed to avoid unnecessary prop drilling)

johbaxter:


johbaxter:
I'm going to ask you to put your own PR, in as I can't make a direct review on a PR I created.  

But below I have my comments.

- [x]  Selected and hovered mask for page currently floats over the panel tab (display: none on StyledTitle if page block)

image

- [x] When adding in a page, can we put a default route to avoid confusion for the user.  (You can make the page-id the route)

- [ ] Modify Layers panel to look like designs. (Section for pages, and section for the selected pages layers)

image

- [x] Interaction when you select a block needs to work with Layers Panel.  If the layers is collapsed, and i select a block on the UI, layers should then open up

johbaxter:
Also noticing interaction with the Link Block is off

When i put a `/` for the destination i get this error

image

We also may want to make things easier so we do not have to force user to add the whole semoss url to the destination.  And rather type something like this in to reference another page in their application

image

image

johbaxter:
Migration function for old pages that have undefined route

PR #455, opened 2025-01-07
feat(client): allow reorder notebook event listeners

### LINKED ISSUE #432: [TASK] Allow Reorder of executed Notebooks tied to all event listeners
### Description 

The event listeners we currently do not allow flexibility of rearranging the notebooks that get executed.  We need to allow draggable behavior to reaarange the notebooks that get executed on event listeners

### Existing

[ListenerSettings.tsx](https://github.com/SEMOSS/semoss-ui/blob/dev/packages/client/src/components/block-settings/ListenerSettings.tsx)

### Reference

[#64](https://github.com/SEMOSS/community/issues/64)

### Tasks

- [x] Allow reorder of executed notebooks on event
- [x] Rename "Run Query" to "Run Notebook" **
- [x] Also show what Notebook is executed right under "Run Query"'

**
image

Paulson-Robert:
Added Draggable feature to the event listeners using react-beautiful-dnd package.


packages/client/src/components/block-settings/ListenerSettings.tsx
```diff
@@ -135,59 +137,160 @@ export const ListenerSettings = observer(
...
+        /**
+         * Reorder columns
+         * @param startDragIndex
+         * @param stopDragIndex
+         */
+        const reorderListeners = (
+            startDragIndex: number,
+            stopDragIndex: number,
+        ) => {
+            const updated = [...listeners[listener]];
```
trevorzylks: Add checks to ensure startDragIndex and stopDragIndex are not out of bounds or are null values.

` if (!startDragIndex || !stopDragIndex || startDragIndex < 0 || stopDragIndex < 0) {
   console.error('Invalid drag indices');
      return;
    }`
    
 The above code may not be entirely accurate according to the library, but we should ensure startDragIndex and stopDragIndex are valid and throw errors as needed.

packages/client/src/components/block-settings/ListenerSettings.tsx
```diff
@@ -135,59 +137,160 @@ export const ListenerSettings = observer(
...
-                                primary={
-                                    <Typography
-                                        variant="body2"
-                                        noWrap={true}
-                                        title={ACTIONS_DISPLAY[a.message]}
+                <DragDropContext
+                    onDragEnd={(result) => {
+                        // ingnore if no destination
+                        if (!result.destination) {
+                            return;
```
trevorzylks: Provide an error alert or a console.log of the issue here. If there is no destination, alert the user. Do we need to check for a null result.source value here as well?

packages/client/src/components/block-settings/ListenerSettings.tsx
```diff
@@ -135,59 +137,160 @@ export const ListenerSettings = observer(
...
+                    }}
+                >
+                    <Droppable droppableId="droppable--list">
+                        {(provided) => (
+                            <List
+                                disablePadding={true}
+                                ref={provided.innerRef}
+                                {...provided.droppableProps}
+                            >
+                                {blockListeners?.map((a, aIdx) => (
```
trevorzylks: `blockListeners && blockListeners.map...`

packages/client/src/components/block-settings/ListenerSettings.tsx
Paulson-Robert: As react-beautiful-dnd package got deprecated in our application, migrated all implementation to dnd-kit(new package)
The above commented bound checks are done in new implementation.

packages/client/src/components/block-settings/ListenerSettings.tsx
Paulson-Robert: As react-beautiful-dnd package got deprecated in our application, migrated all implementation to dnd-kit(new package)
The above commented bound checks are done in new implementation.

packages/client/src/components/block-settings/ListenerSettings.tsx
Paulson-Robert: As react-beautiful-dnd package got deprecated in our application, migrated all implementation to dnd-kit(new package)
The above commented null checks are done in new implementation.

Paulson-Robert:
I found the usage of the react-beautiful-dnd package in two other places. Due to the package deprecation, the drag-and-drop feature was not working there either. Therefore, I migrated those sections as well:
**Files**
1. Grid block - GridBlockColumnSettings.tsx
2. Toggle options block - OptionsSettings.tsx

johbaxter:
@AAfghahi  will review

PR #490, opened 2025-01-23
fix(ui): change showOrDivider to a function

### LINKED ISSUE #480: [BUG] When no OAuth login method is added, we shouldn't show the -----or------ on the login page
### Description 

On the login page, you will see the ------or------- and then its empty under. We should just now show the divider at all.

Image

boredcod:
Fixed the showOrDriver by changing it to a function. In JS, -1 is true

### DIFF
--- a/packages/client/src/pages/LoginPage.tsx
+++ b/packages/client/src/pages/LoginPage.tsx
@@ -494,14 +494,33 @@ export const LoginPage = observer(() => {
 
     // get the proviers
     const providers = [...configStore.store.config.providers];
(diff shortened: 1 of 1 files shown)


neelneelneel:
@themaherkhalil - Is there an easy way to enable all providers?

phkem (approved):
Overall looks good! Only possible change is if the searchValues are used anywhere else in the application, then we may want to consider moving this array to a constants file. But if its only used on login, then its probably fine.

PR #495, opened 2025-01-27
feat(client): permission request logic cleanups

### LINKED ISSUE #121: [TASK] Clean up Permissions Request Logic
While user is not in "ADMIN MODE": 

**1. The Change Access Modal:**
- If an access permission already exists for the user (not requesting access for first time), clicking "change access" to request elevated access brings up the following form. This form should indicate what level of status they currently have, which it currently does not. 
image

**2. Members Table Component**
- If user has Editor level permission, they should not be able to change anyone's permissions to Author, including themselves. For example, if they go to the Settings page, they should only be able to click the radio button options for "Editor" and "Read-Only" when adjusting permissions for existing members (including themselves) or adding a new member. 
- If the user has Read-Only level permission, they should not be able to change anyone's permissions to Author or Editor, including themselves. Question: when are Read-Only users able to view the Members Table? I thought they cannot access it through the App Settings page, since that's in the Edit view. 
- For both above scenarios, the radio buttons that the user cannot modify should be disabled. 

Image

**3. Add Member Modal:**
- If user has Editor level permission, they should not be able to add anyone as an Author (including themselves). 
- If the user has Read-Only level permission, they should not be able to add anyone as an Author or Editor (including themselves). 
- For both above scenarios, the unavailable options should not appear on the modal to the user at all. 

Image

**4. Account for Admin Only Permissions Info in Config Fields**
- The permissions logic needs to account for the following config fields that state whether certain functionality should be limited to only Admins
Image

Paulson-Robert:


johbaxter:
@antgibson96 will review

packages/client/src/components/settings/MembersAddOverlay.tsx
```diff
@@ -73,6 +73,18 @@ const permissionMapper = {
     'Read-Only': 'READ_ONLY', // DISPLAY: BE
 };
 
+const accessPriority = {
```
antgibson96: let's use the mapping directly above this, seems similar enough.

packages/client/src/components/settings/MembersTable.tsx
```diff
@@ -176,6 +177,18 @@ const permissionMapper = {
     'Read-Only': 'READ_ONLY', // DISPLAY: BE
 };
 
+const accessPriority = {
```
antgibson96: pull whatever you resolve for these two mappings into a utility function to be used across the app.

Paulson-Robert: Made it as utility function

johbaxter: @antgibson96 Can you take another pass at this when you get the chance.

antgibson96 (approved):

ehynd:
@johbaxter - it looks like this was reviewed and approved feb 5 but never merged in / closed. is it good to close?

johbaxter:
@AAfghahi when you review please don't just merge if you approve the changes.  

This is pretty critical and touches settings on everything, give either @ehynd, myself or the Semoss Dev chat a 2hr warning before we merge

AAfghahi:
I reviewed this and there was nothing breaking. However, I need to change my social properties and check.

ehynd:
@AAfghahi did you get the chance to check w social properties changed

PR #499, opened 2025-01-28
Task 111: added title elispse after two lines and tooltip hover feature

abarricks:
## Description
Enhanced the Typography component to improve readability of long text in body1 variant by adding automatic truncation after two lines with an ellipsis. Added a smart tooltip that appears only when text is actually truncated.

## Impact Areas
This change affects all components using Typography with body1 variant, including similar items on other pages (e.g. list of features)

## Screenshots
image


neelneelneel (dismissed):
Instead of overloading Typography, can you create a new component with a clip/tooltip feature. We can then be intentional of when we want to use it vs not.

antgibson96 (approved):

PR #510, opened 2025-01-29
fix(ui/client): bug fix for multiple calls

### LINKED ISSUE #482: [BUG] FE makes duplicate calls on app loading
### Description 

On the login page, it will call /config twice

Image

If you are already logged in and refresh the page, it will call twice:

/config 
/runPixel with GetUserInfo();
auth/admin/user/isAdminUser

Image

trevorzylks:
Bug fix for multiple calls being made to /config, getUserInfo, and isAdminUser. I believe this may have been a re-rendering "issue" caused by the context provider. I used setState to handle the logic for running the initialize function. 

**set env variable > setState for configuration > run configuration once**

### DIFF
--- a/packages/client/src/App.tsx
+++ b/packages/client/src/App.tsx
@@ -68,11 +68,7 @@ export const App = () => {
             // noop
         }
 
-        if (!configuration) {
-            setConfiguration(true);
-        } else {
-            setConfiguration(false);
-        }
+        setConfiguration(true);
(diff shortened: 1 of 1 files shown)


neelneelneel:
We wrap code in strict mode. This is by design in dev. See https://18.react.dev/reference/react/StrictMode#fixing-bugs-found-by-double-rendering-in-development

PR #511, opened 2025-01-30
fix(ui/client): bug fix for the help button as the menu was not visible

mananmittal07:
The help button menu section was commented out so we have uncommented it and raised the pr so that the functionality of the help button is up and running


neelneelneel:
Help pulls from the Themes DB. It shouldn't be hardcoded.

PR #512, opened 2025-01-30
feat(client): improve styling of switch components in block settings

### LINKED ISSUE #486: Improve styling of switch components in Block Settings
### Description 

For switch components within Block Settings, we need to allow the label next to it to display within 16 px of the switch. If there is overflow, then wrap and make sure the container with both the label and the switch is middle-aligned from top to bottom. Additionally, we should use the "small" switch instead of the current one, which looks oversized amongst our other components 

### Existing

Below are some screenshots of examples where these switch components are found in:

Screenshot 2025-01-09 at 11 38 25 AM 

Screenshot 2025-01-09 at 11 41 20 AM

Image

Image

Image

Image

Image

Image

### Reference

This item was carried over from this ticket, which has been closed: https://github.com/SEMOSS/semoss-ui/issues/460

### Tasks

For all blocks with switch components:
- [ ] Allow the switch labels to display within 16 px of the switch. 
- [ ] In case of overflow, wrap and ensure container with label and switch is middle-aligned from top to bottom. 
- [ ] Swap the current switch component for the "small" switch.

NayanKumar809:
After changes, the switch button looks like this:
image

Before changes, it was looking like this:
image

### DIFF
--- a/libs/ui/src/components/Switch/Switch.tsx
+++ b/libs/ui/src/components/Switch/Switch.tsx
@@ -2,10 +2,6 @@ import { ReactNode } from "react";
 import { Switch as MuiSwitch, SxProps, styled } from "@mui/material";
 
 const StyledSwitch = styled(MuiSwitch)(({ theme }) => ({
-    width: "52px",
-    height: "32px",
-    padding: 0,
-
     "& .MuiSwitch-switchBase": {
         padding: 0,
         margin: "4px",
@@ -33,8 +29,6 @@ const StyledSwitch = styled(MuiSwitch)(({ theme }) => ({
 
     "& .MuiSwitch-thumb": {
         boxSizing: "border-box",
-        width: "24px",
-        height: "24px",
         color: theme.palette.background.paper,
     },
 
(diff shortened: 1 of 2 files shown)


neelneelneel:
@natdink Thoughts? This doesn't match the spec

natdink:
> @natdink Thoughts? This doesn't match the spec

Agree, not sure what the original request was but as a team we had decided to use the "apple" styling of the switch component. I think if anything we need to work on the styling of the existing switch component to provide sizing options and scale appropriately.

I know we aren't on Material 3 yet but you should be able to utilize their specs on the switch component: https://m3.material.io/components/switch/overview

johbaxter:
Im giving this a pass. 

Nayan added custom styles on the size property for switch, nothing bad pushed in just a smaller version of what we already have that represents an apple switch.

I'm not seeing anything the material 3 link from above specifying multiple sizes of the switch, no visual indicators.  But as for the normal size of switch that has not been affected.  

If this is a big issue on the styling of the switch we can bring this up in a new ticket.  Looking at styles im good with it, if you see anything please point out the exact pixel we are messing up

PR #516, opened 2025-01-30
Resolve: "[TASK] Implement frontend for user limits on model usage"

### LINKED ISSUE #475: [TASK] Implement frontend for user limits on model usage
### Description 

The backend has been merged/pushed: https://github.com/SEMOSS/Monolith/pull/70 it is new parameters in the existing /addEnginePermission and /editEnginePermission endpoints the FE is performing

The additional parameters are:

usageRestriction ('token' or 'compute' as the values)
usageFrequency ('DAY', 'WEEK' or 'MONTH')
maxTokens
maxResponseTime

### Reference

https://github.com/SEMOSS/community/issues/79

JoshRome26:
Resolved in adding in the model limit features: 

Catalog Model: 
image
image
image

Member's Page: 
image
image

### DIFF
--- a/packages/client/src/components/settings/MembersAddOverlay.tsx
+++ b/packages/client/src/components/settings/MembersAddOverlay.tsx
@@ -789,6 +789,7 @@ export const MembersAddOverlay = (props: MembersAddOverlayProps) => {
                                 <TextField
                                     label="Max Tokens"
                                     value={maxTokens}
+                                    type="number"
                                     onChange={(e) => {
                                         setMaxTokens(e.target.value);
                                     }}
@@ -799,6 +800,7 @@ export const MembersAddOverlay = (props: MembersAddOverlayProps) => {
                                     <TextField
                                         label="Max Response Time"
                                         value={maxTime}
+                                        type="number"
                                         onChange={(e) => {
                                             setMaxTime(e.target.value);
                                         }}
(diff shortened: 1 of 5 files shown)


JoshRome26:
I believe there is an issue with the endpoint '/api/auth/user/getAllUsers' as the registerUser / editUser seem to accept the data being submitted, but the get never returns the data submitted.

johbaxter:
Pushed and Fixed bugs @JoshRome26 
 
**Engine Settings page**

- [ ] Fixed editting member permissions on the table so the model usage persists
- [ ] Add and Edit member modal, switched Max Token and Compute Time to number picker (prevent user error)
- [ ] Fixed Edit member modal, able to update permission level and model usage
- [ ] FYI: If user doesn't fill out any of the model usage, we do not send the keys in the map that gets sent
 
**Member Settings Page**

- [ ] undefined values do not get sent in params on register
- [ ] Add and Edit member modal, switched Max Token and Compute Time to number picker (prevent user error)
- [ ] Add Member was not working, the keys were being sent snake_case instead of camelCase 
 
Notes for you @JoshRome26 . extra testing please as not fun as it sounds but avoids back and forth, also we have access to the backend code so we can go look at how these keys are being expected.  Search for MonolithApplication, and then ctrl click into the one that sounds most like what your work is associated to.
 
Sidenote: The way we send these keys versus the way receive is inconsistent, can fix at a later point in time (BE) @themaherkhalil 
 
---------------------------------------------------------------------------------------------------
 
Tickets that we need to add: 
[ ] UX on table with multiple columns
[ ] URGENT pagination on members table (will be in next sprint) @kyang4 

image

PR #534, opened 2025-02-04
483 task new UI to work with csrf

### LINKED ISSUE #483: [TASK] New UI To Work with CSRF
### Description 

This is a security setting. When the /config call returns "csrf":true, the FE needs to make an additional call before each POST message. In the screenshot below, you can see on the new UI, it tries to make the POST message directly and gets a 403

Image

Image

On the legacy BI UI, the FE makes the call to /config/fetchCsrf  and gets back a header X-Csrf-Token: <random ID". 

Image

This X-Csrf-Token must then be used for all subsequent POST requests

Image

### Existing

To test this locally, you must turn on the CSRF filter in Monolith project web.xml

Image

memisrose:
using axios request interceptor to fetch and set x-csrf-token header


antgibson96 (approved):

neelneelneel (changes requested):
fetchCsrf is still called when csrf is not enabled. 

image

Hint: Take a look at the CSRF in sdk

neelneelneel (dismissed):
fetchCsrf is called multiple times. This is with StrictMode turned off. See:

image

PR #564, opened 2025-02-11
Resolve: 'Fix pagination on Members/Users table'

### LINKED ISSUE #561: [BUG] Fix pagination on Members/Users  table

The pagination feature on the "Users" table within Member Settings is broken - the table does not accurately display the true count of the table and does not allow users to click through to subsequent pages of the table. 

1. Go to Settings and toggle Admin on 
2. Go to Member Settings
3. See the "Users" table and observe that it displays that there are only 5 users in the system and you cannot click to the next page to see more. However, there are more than 5 users because the search bar still works and you can search for users who do not appear in the five listed. 
4. Try searching for a member not listed and see that they exist within the table. 
5. Note that changing the "Rows per page" from 5 to 20 allows 20 users to appear, but still does not capture the true count and does not fix the pagination problem. 

### Expected results

The number of users displayed is accurate, and if it exceeds the "Rows per page" value, the pagination feature in the bottom right allows you to click through to subsequent pages to view all the users. 

### Actual results

If the number of users in the table exceeds the "Rows per page" value, only that number of users will be displayed and there is no way to click through to subsequent pages. Also, the number of users displayed in the top left is just the "Rows per page" value instead of the true count. 

#### Screenshots

Image
(personal information redacted)

### Environment

- browser type and version:
- node.js version: `node -v`

### Additional context

Where are all the cases this bug occurs in the app? Doesn't seem to be a problem when viewing Members table for a specific engine, but does appear for "Users" table under "Member Settings" - need to verify.

### Tasks

- [ ] Identify all locations where this table pagination bug occurs in the app
- [ ]

JoshRome26:
Closes #561


johbaxter:
Ty sir @JoshRome26 

Looks good awaiting backend merges on below:
https://github.com/SEMOSS/Monolith/pull/95
https://github.com/SEMOSS/Semoss/pull/448

Im ready to merge this in once those are in

johbaxter (approved):

PR #570, opened 2025-02-12
feat(client): typography block enhancements

### LINKED ISSUE #384: [TASK] Typography Blocks - Add floating menu when a typography block is selected in the canvas
### Description 

Add a floating menu that allows the user to change the font style and weight for typography blocks when the block is selected in the canvas. They can still change font and weight in the block settings but this is a way for users to easily update these common items quickly. Since this menu now appears, it may conflict with the duplicate/delete menu option that appears when a block is selected. Please move it to the top right of the box as shown in the Figma. Notice the change to use icon buttons only instead of text buttons.

### Existing

Current state of the duplicate/delete buttons is in the middle of the page near the container: 
Image

### Reference

https://www.figma.com/proto/Qj4tWPsniFOJ6qSLfxOaBT/UI-Builder---Design?node-id=8026-105361&t=cJ8UlqlGxxDtWQec-1

### Tasks

Image

- [ ] Move duplicate and delete button menu to the top right and update to icon buttons
- [ ] Create hover action for each respective icon button to say "duplicate" and "delete"
- [ ] Create font style and weight input menu under typography blocks when selected

Paulson-Robert:


johbaxter (approved):
Looks good Paulson ty

PR #602, opened 2025-02-19
Feature/agent builder flow enhancements

### LINKED ISSUE #485: [TASK] Genericize inputs on agent builder app flow
### Description 

When a user selects words/phrases from the prompt to become inputs in an agent builder app, the associated labels should ideally be automatically genericized from the specific examples in the original prompt. 

For example, if the initial prompt is "I am a chef looking for a recipe with lamb, tomato, and rigatoni" and "lamb", tomato", and "rigatoni" become input fields, they should be genericized to something like "protein", "vegetable", "grain" or "ingredient 1", "ingredient 2", "ingredient 3". The user can then go in and modify if needed to fit their specific intention, but the genericized labels would be a good starting point and less confusing that the current setup. 

### Reference

https://github.com/SEMOSS/community/issues/48

22mayank:


Srk15:
Closing PR as functionality not required for now.

PR #615, opened 2025-02-21
feat(client): Added meta fields and implemented SetProjectMetaData on NewAppModal

### LINKED ISSUE #628: Update Create App experience and make small updates to app details

The app details page lacks content this is due to the users not needing/are not required to input content about their apps when they start.

Update the create new app modal to have additional (required) fields like description, tags, markdown etc. Use tooltip for markdown to give them helpful ways to use it.

 _**Side note:**_

On app details page, dependencies should not be manually entered on the UI if its a drag and drop (type) app.

Instead, dependencies should be pulled from the blocks.json. 

For code apps, we will show a manual option to add dependencies on the UI and can possibly set up a step in the SDK to put in a file with acceptable key values. ( this should work already as it stands )

### Tasks

A list of individual tasks / issues that need to be completed before the ticket is resolved

- [ ]
- [ ]

j-arias-dev:
- Added description and tag meta fields to NewAppModal. 
- Implemented setProjectMetadata on submit in NewAppModal.

- <s>For NewAppModal: Waiting to find out how long it will take to update the CreateAppFromBlocks and CreateProject reactors to pass the metadata."</s>

**NewAppModal**
image

image


johbaxter (commented):
image

Is there a need to call SetProjectMeta, if we dont provide description and tags

j-arias-dev:
@johbaxter  **implemented a metadata check in order to run SetProjectMetadata**

image

PR #617, opened 2025-02-25
433 reordering layers panel

### LINKED ISSUE #433: Allow Reordering of layers through panel
### Description 

On our drag and drop apps we need to be able to offer extra flexibility to allow reordering of blocks through the layers panel, instead of being forced to reorganize items on the Designer panel.

**Rules**

Page should not be moveable

### Existing

[LayersPanel.tsx](https://github.com/SEMOSS/semoss-ui/blob/dev/packages/client/src/components/blocks-workspace/panels/LayersPanel.tsx)

### Reference

image

### Tasks

- [ ] Allow the ability to reorder blocks on layers panel

Shubham009-prog:
Added the feature of layers reordering using DND kit library.


johbaxter (approved):
Functionality is good, will have to follow up with a style ticket to address scrolling with a bunch of containers

PR #648, opened 2025-03-03
build(admin-query): admin-query table changes

SuryaNarayanan-Kgs:


antgibson96 (approved):

packages/client/src/pages/settings/AdminQueryPage.tsx
```diff
@@ -1,19 +1,22 @@
...
     TextArea,
 } from '@semoss/ui';
-
+import {
+    Table,
+    TableHead,
+    TableBody,
+    TableRow,
+    TableCell,
+} from '@mui/material';
```
neelneelneel: Use components from @semoss/ui. Extend those if necessary

neelneelneel (dismissed):

packages/client/src/pages/settings/AdminQueryPage.tsx
johbaxter: addressed

ppatel9703: @SuryaNarayanan-Kgs Moving forward can we focus on using pre existing components inside of semoss/ui

PR #650, opened 2025-03-03
Escape special chars in get requests

phkem:


themaherkhalil:
Done in following PR: https://github.com/SEMOSS/semoss-ui/pull/651

PR #658, opened 2025-03-04
Vitest Unit Testing 

memisrose:
Adding unit testing into @semoss/client to first set up the testing env. 

Tests can run with `pnpm run test` but will also run with pre-commit. Only test files that have been added or changed will be tested.

image

An example of a test failing and exiting the pre-commit:
image


johbaxter:
image

Can we get a script in the root package.json too

johbaxter (approved):

memisrose:
PR for adding script to root: https://github.com/SEMOSS/semoss-ui/pull/693

PR #664, opened 2025-03-05
Origin/usage report page

### LINKED ISSUE #675: [FE] Add Engine Usage Tab under Engine Details Page
We want to be able to have a new table for Models that display two seprate sets of information. 

The first table being Engine(model) usage by user. The table should be able to filter on user and should be able to handle pagination 

The second table should display Engine(model) usage by project. The table should be able to filter on project and should be able to handle pagination.

SuryaNarayanan-Kgs:
## How to Test

2. Expected outcomes


johbaxter:
image

PR #682, opened 2025-03-08
584 create system popover

### LINKED ISSUE #584: Create System Popover Block
### Description 

I envision an important setting on the Popover Block for:

Anchor Block --> data.anchorEl

The options for this setting will be a list of all blocks other than the page-block.

It is safe to assume that this Popover will only be open on hover or onclick events 

Outside of this setting, i envision the settings will have to be similar to the modal block so you can edit it while working in the designer.

The content of this popover can use our `<Slot />` component similar to our Modal

I can explain further if needed

### Existing

This is a new Block

### Requirements

- [ ]
- [ ]

Shubham009-prog:
## Description
Added a new Block (Popover Block).

## Changes Made
1. Added popover images
2. Added popover in default menu.
3. Added a new folder for popover.

## How to Test
1. Just Drag and drop the Popover Block to designer page.
2. Also can customize the popup appearence.


johbaxter (approved):
Looks good enough for first cut.

I commented out Hover trigger, does not work.

I think it would be cool to have some load state on this where we just overlay loading indicator in popover window

Maybe this can be a setting on a Container block

PR #689, opened 2025-03-11
feat(client): adds a switch block component

### LINKED ISSUE #594: Switch Block component
### Description 

We want to add a component/block to the System Blocks of Drag and Drop that allows users to select a value using a slider.

Image

For inspiration: https://mui.com/material-ui/react-switch/

### Existing

We do have a Toggle Button block currently:

Image

But a Switch Toggle might be more applicable/intuitive for certain use cases. 

### Requirements

- [ ]
- [ ]

22mayank:
## How to Test

2. Expected outcomes


johbaxter (approved):
Cool for now

PR #722, opened 2025-03-18
feat(libs): query settings input tool enhancement

### LINKED ISSUE #638: QueryInputSettings - show all blocks/notebooks/cells in list even if not variable
### Description 

We need to fix the Settings Component that allows us to bind to variables on our Designer

The Menu currently displays block, notebook, and notebook cell variables.  But there is not ability to bind to something that has not been made a variable yet.

I'm proposing having changes to the menu, for certain sections.  

- Block variables
- Notebook and cell variables
- Things that have not yet been made a variable

Modifications will be made to this component: `QueryInputSettings.tsx`

Image

This is a very lofi mockup of what i envision with this menu

Image

When a user clicks the menu item for something that has not yet been made a variable, we should streamline the process and automatically create the variable for them and add it to the `state.variables` portion of our JSON. 

For blocks, the default variable name can be the name of the block. 

Other considerations for future:
- We need a way of notifying user that a new variable has been created (not obvious if they do not have Variables panel open)
- Using hints to simplify the dropdown menu while still allowing users to select what they need
- We need to replace the "Query" placeholder text with some thing more accurate, since the inputs are not limited to just queries.

### Existing

`QueryInputSettings.tsx`

### Requirements

- [ ] Mockup designs for said menu
- [ ] Make modifications to QueryInputSettings.tsx to have the menu item handle multiple sections
- [ ] Stretch Goal allow VS code type hinting for properties like `.isError`, `.isLoading`, and etc...

Paulson-Robert:
## How to Test

2. Expected outcomes


johbaxter:
https://github.com/SEMOSS/semoss-ui/pull/750

Merged into here

PR #752, opened 2025-03-24
Client components and Mermaid variations

### LINKED ISSUE #654: Migrate Mermaid variants to Client Blocks
### Description 

Under Client Blocks, we want to include the existing built-out MermaidJS variants. Each diagramming method should be its own Client Block the associated picture should be the same one that it was previously. 

### Existing

These Blocks already exist, we just have to move them to Client Blocks and ensure they have the correct image attached. 

### Requirements

- [ ]
- [ ]

Shubham009-prog:
## Description

- Added client block tab in Blocks
- Added logic to group client block


johbaxter:
Alright so it looks like the way the AddBlocksMenuCard is currently setup.  We will not need a seperate component for Adding in a client component.  

So you can get rid of the custom implementation `AddClientBlocksMenuCard.tsx`.

I have added 2 client components.  One, which is one of the variations of this Mermaid block, and another with a container with an input as a child of the container.

I need you to clean up the code `BlocksMenuPanel.tsx` where it the items flow through the `renderedItems` as you had before. But also need to be able to work with custom sectioning as i have in my screenshot below.

image

One of the grouped components looks like this

image

We also need you to add the rest of the mermaid variations in the list for get client blocks

johbaxter:
Follow up tickets will be to incorporate AddBlocks Reactor.

johbaxter:
Going to address [Duplicate Bug](https://github.com/SEMOSS/semoss-ui/issues/781) in this same PR

@Paulson-Robert

johbaxter:
https://github.com/SEMOSS/semoss-ui/pull/786

Code from this folded into this PR

Ty @Paulson-Robert

PR #785, opened 2025-04-03
Iterator Block

### LINKED ISSUE #692: Iterator Block thought process
In Drag & Drop apps, we need a way to conditionally render blocks based on elements in an array. For example, if we have a variable that's a list of names, we may want to display each name in its own text field. Extending the example, we need a way in Drag & Drop to duplicate the following JS code:
```
const arr = ["Person A", "Person B", ..., "Person Z"];

return (
  <>
    {arr.map((el, index) => (
      <div key={el}>{`${index}: ${el}`}</div>
    ))}
  </>
);
```

Baxter's and my idea is to create and Iterator Block that wraps any elements that need to be duplicated. At minimum, the Iterator Block should allow the user to select which array over which they would like to iterate. In pseudocode:
```
arr = ["Person A", "Person B", ..., "Person Z"];

IteratorBlock: {
  array: arr; // the array over which to iterate
  children: {
    // the contents that should render once per # of array entries
  }
}
```

Next comes the question of how the child elements should index into the array to get their specific entry. My current thought is that we can create a new variable type, Iterator. Then, in pseudocode, we can do something like
```
arr = ["Person A", "Person B", ..., "Person Z"];
i = new Iterator;

IteratorBlock: {
  array: arr; // the array over which to iterate
  iterator: i; // the variable used for iteration
  children: {
    // the contents that should render once per # of array entries
    TextBlock: {
      text: arr[i] // use the Iterator variable to index into our list
     }
  }
}
```

NaveenRamasamy028:
## Description
Created an Iterator Block So that the blocks that are dropped inside iterator block can be iterated


johbaxter:
[iterator_blocks.json](https://github.com/user-attachments/files/19594146/iterator_blocks.json)

Demo Iterator Block Template

@NaveenRamasamy028 can you try to build with this array that i have there

johbaxter:
Closed with another PR

PR #825, opened 2025-04-11
feat(client): open new page on creation from layers panel

### LINKED ISSUE #283: Open new page upon creation from the Layers panel

Currently, when you add new page from the Layers panel, the new Page is created but must be selected from the menu of pages to be opened in a new tab.

Perhaps it would be more intuitive if the new pages is opened upon creation. So when a user adds a new Page from the Layers panel, a new tab opens with that new Page and becomes the selected/visible page. This would help confirm the new Page's creation by bringing attention to it while also potentially reducing a click, since if a User is creating a new Page, it is likely they will modify it next.

### Describe alternatives you've considered

The current is the alternative. If Users are creating a lot of pages at once but not immediately modifying them, it might be annoying for them to automatically open - so perhaps we can do more user research into the typical work flow before implementing a change.

### Additional context

Proposed by @sarahtrans 

Current state:
Image

Image

### Requirements

- [ ]
- [ ]

Paulson-Robert:
## How to Test

2. Expected outcomes


johbaxter:
Closed with #895 

Ty Paulson

PR #909, opened 2025-04-17
test(client): add tests for Markdown Block

stelbailey:
## Description
Added tests for the Markdown Block

1. Checks that it renders correctly
2. Checks the content in the markdown is correct
3. Checks the content in the markdown is correct _after_ animation (streaming)
4. Checks styles are applied to the markdown lbock.

## How to Test

5. Expected outcomes


memisrose (changes requested):
Can we change "test" to "it"? Will add corresponding notes of test vs it inside the testing ReadMe but otherwise this looks great!

libs/renderer/src/testing/block-defaults-testing/MarkdownBlock.spec.tsx
```diff
@@ -0,0 +1,74 @@
+import { render, screen, waitFor } from "../utils";
+import { expect, test } from "vitest";
+import "@testing-library/jest-dom";
+
+import { MarkdownBlock } from "../../components/block-defaults/markdown-block/MarkdownBlock";
```
memisrose, line 5: please use relative imports ie '@/components...'

memisrose (changes requested):

Issue #734, opened 2025-03-19
Radio Block


memisrose:
### Feature:
Block Defaults

### Component / Function:
The specific component this test will cover.

### Given:

![Image](https://github.com/user-attachments/assets/48e132b0-620d-4993-8ca4-c87ee0bfb906)

### When:
When an event or user action occurs

### Then:
The expected outcome of this test

PR #918, opened 2025-04-18
test(ui): add radio block test

alaleung:
## Description
test(ui): add radio block test
https://github.com/SEMOSS/semoss-ui/issues/734


memisrose (approved):

libs/renderer/src/testing/block-defaults/RadioBlock.spec.tsx
```diff
@@ -0,0 +1,127 @@
+import { render, screen } from "../utils";
+import { expect } from "vitest";
+import "@testing-library/jest-dom";
+
+import { RadioBlock } from "../../components/block-defaults/radio-block/RadioBlock";
```
memisrose: please use relative imports ie '@/components...'

memisrose (changes requested):

memisrose (approved):

Issue #719, opened 2025-03-18
Logs Block


memisrose:
### Feature:
The overarching feature this component test is apart of.

### Component / Function:
The specific component this test will cover.

### Given:
Preconditions or current state

### When:
When an event or user action occurs

### Then:
The expected outcome of this test

PR #919, opened 2025-04-18
test(ui): add logs block test

alaleung:
## Description
test(ui): add logs block test
https://github.com/SEMOSS/semoss-ui/issues/719

### DIFF
--- a/libs/renderer/src/testing/block-defaults/LogsBlock.spec.tsx
+++ b/libs/renderer/src/testing/block-defaults/LogsBlock.spec.tsx
@@ -2,7 +2,7 @@ import { render, screen } from "../utils";
 import { expect } from "vitest";
 import "@testing-library/jest-dom";
 
-import { LogsBlock } from "../../components/block-defaults/logs-block/LogsBlock";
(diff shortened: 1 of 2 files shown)


libs/renderer/src/testing/utils/index.tsx
```diff
@@ -1,20 +1,26 @@
 import "@testing-library/jest-dom";
+import { cleanup } from "@testing-library/react";
```
memisrose: remove unused import

memisrose (changes requested):

libs/renderer/src/testing/block-defaults/LogsBlock.spec.tsx
```diff
@@ -0,0 +1,72 @@
+import { render, screen } from "../utils";
+import { expect } from "vitest";
+import "@testing-library/jest-dom";
+
+import { LogsBlock } from "../../components/block-defaults/logs-block/LogsBlock";
```
memisrose: please use relative imports ie '@/components...'

memisrose (changes requested):

memisrose (approved):

PR #921, opened 2025-04-20
686 fix styles on dropdown menu items within team add project engine and members table

### LINKED ISSUE #686: Fix styles on Dropdown Menu Items within Team Add Project, Engine, and Members table
### Description 

Upon adding the infinite scroll ticket in, I noticed the Menu items for those same dropdowns lack clarity with design.  Can we fix the styles for all the dropdowns in each of the tables

Add Project
Add Engine
Add Members

### Existing

Image
Image

### Requirements

j-arias-dev:
## Changes Made

- Changed the state of Members, Projects, and Engines to object types to fix bugs caused by duplicated object ID keys and incorrect type key.
- One thing I noticed while implementing type objects into the team permissions page is that the `getProjectsForGroup` function returns two duplicate keys. Both keys return the same value, but one is `project_id` while the other is `projectid`. Similarly, for the `getEnginesForGroup` function, one key is `engine_id` while the other is `engineid`. In my PR, I updated it to use the one with the underscore.
- Previously we were not passing a type value to the `editEnginePermission` reactor because the engine object lacked a `type` key. I have updated it to pass the correct key, which is `engine_key`.
- Modified the dropdown styling in the Autocomplete component for Members, Projects, and Engines.

## How to Test

Test the updated dropdown styling options for the Add Members, Add Projects, and Add Engines modals.
Test CRUD functionality for Members, Projects and Engine
2. Expected outcomes

### Add Members
image

image

### Add Projects
image

image

### Add Engines
image

image


packages/client/src/stores/monolith/monolith.store.ts
```diff
@@ -1274,11 +1274,7 @@ export class MonolithStore {
      * @param userId
      * @returns
      */
-    async deleteTeamUser(user: {
-        groupid: string;
-        type: string;
-        userid: string;
-    }) {
+    async deleteTeamUser(user) {
```
AAfghahi: Why get rid of the object here? Do we have separate versions of user?

packages/client/src/components/teams/TeamEnginesTable.tsx
```diff
@@ -732,25 +771,26 @@ export const TeamEnginesTable = (props: EnginesTableProps) => {
...
                                                             </Stack>
                                                         </NameIDWrapper>
                                                     </UserInfoTableCell>
                                                     <Table.Cell size="medium">
                                                         <RadioGroup
                                                             row
-                                                            defaultValue={
-                                                                engine.permission
-                                                            }
+                                                            defaultValue={String(
```
AAfghahi, line 786: The default value is a number for permission, no?

AAfghahi:
It looks good, some small comments, going to test it in the UI tomorrow.

packages/client/src/components/teams/TeamEnginesTable.tsx
j-arias-dev: @AAfghahi 
Correct, but the value and default value of the Radio component are only of type string.

packages/client/src/stores/monolith/monolith.store.ts
j-arias-dev: @AAfghahi  
Now that we have implemented object types, we can pass the object directly instead of creating a user object in the parameters. To prevent adding various object types to the monolith store file, I can pass only the necessary values into the parameters rather than creating or passing an object. Will that be better?

AAfghahi (approved):

AAfghahi (changes requested):
Needs this BE fix: https://github.com/orgs/SEMOSS/projects/34/views/2?pane=issue&itemId=107664973&issue=SEMOSS%7CSemoss%7C726

packages/client/src/stores/monolith/monolith.store.ts
neelneelneel: The method will need to be typed and any passed in types must match that.

neelneelneel (changes requested):

packages/client/src/stores/monolith/monolith.store.ts
j-arias-dev: @AAfghahi  @neelneelneel 
The deleteTeamUser change has been pushed. The branch has been updated.

neelneelneel (approved):

AAfghahi (approved):

packages/client/src/components/teams/TeamEnginesTable.tsx
```diff
@@ -732,25 +771,26 @@ export const TeamEnginesTable = (props: EnginesTableProps) => {
...
                                                                 );
                                                                 updateSelectedEngines(
                                                                     {
-                                                                        engineid:
-                                                                            engine.engineid,
-                                                                        type: engine.type,
+                                                                        engine_id:
+                                                                            engine.engine_id,
+                                                                        engine_type:
+                                                                            engine.engine_type,
```
AAfghahi: Went through this with the BE engineer, and engine_type is not necessary for this particular call. What this is doing is adding an engine to a specific Group. It's just creating a relational database btwn the two tables, so we will still have access to the engine_type. But passing in the engine type, I don't think is necessary, especially since the engine table is type agnostic. 

@anurag91jain  do you have naything to add from our discussion? 

If you delete this portion @j-arias-dev  I think this is good to merge.

j-arias-dev: @AAfghahi  @anurag91jain 
I deleted the `engine_type` from `updateSelectedEngines` and restored the original `editEnginePermission` parameters.
The update has been pushed.

AAfghahi (approved):

AAfghahi:
please resolve the conflicts!

j-arias-dev:
@AAfghahi  
Merge conflicts have been resolved and updates have been pushed.

johbaxter:
Are these changes still relevant, and are we waiting on backend PR to be merged for this too.

@j-arias-dev @AAfghahi

j-arias-dev:
@johbaxter @AAfghahi 
We decided not to pass engine_type into editEnginePermission, so everything is set on my end.
https://github.com/SEMOSS/semoss-ui/pull/921#discussion_r2082141416

johbaxter:
Closing as stale

PR #963, opened 2025-04-22
feat(client): add shared insight id for panels in workspace

### LINKED ISSUE #958: Integrate insight id sharing between workspace panels
### Description 

Shared InsightId to all blocks and code apps will be done through the workspace store

I'll initialize the insightId to the workspace.store and pass it to the all the panels that should have access to it

Add Property to workspace.store for a new insightId, runPixel("1+1", "new) if not there already

Pull the insightId from within each panel.
Ex: TerminalPanel.tsx will need to send insightId in the runPixel call there
** Do to the other Panels DesignerPanel, FileExplorerPanel, FileEditorPanel, ChatPanel

We want to be able to  access the runPixel from workspace through a hook

### Existing

### Requirements

- [ ]
- [ ]

johbaxter:
## How to Test

2. Expected outcomes

### DIFF
--- a/libs/renderer/src/Renderer.tsx
+++ b/libs/renderer/src/Renderer.tsx
@@ -182,7 +182,8 @@ export const Renderer = observer((props: RendererProps) => {
                     background: "rgba(255, 255, 255, 0.5)",
                     zIndex: 1501,
                     position: "relative",
-                    width: "100%"
+                    width: "100%",
+                    height: "100%"
                 }}
             >
                 <Stack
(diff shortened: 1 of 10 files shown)


johbaxter:
Things like FileEditor shouldn't have useWorkspace hook in it.  To avoid the dependency of needing to wrap in Workspace.Context.

Fix Panel components to pass callback function as a prop to child components to avoid having to `runPixel from @api`  but rather call it from `workspace.runWorkspacePixel`

PR #966, opened 2025-04-23
402 task update grid block

### LINKED ISSUE #402: [TASK] Update Grid Block
### Description 

Increase functionality of existing grid block in new UI.  Would suggest starting with seeing how the grid functions in legacy, we want to replicate some of that and change some of that 

### Existing

Current grid ability within new UI: 
- [ ] Select frame and columns
- [ ] Toggle pagination on/off

### Tasks

The following updates should be added to the grid functionality around styling and events: 

Events
- [x] Click event 

Styling 
- [x] Header background color
- [x] Header font color/size 
- [x] Cell font color/size 
- [x] Color by value (should apply to one cell with ability to checkbox to apply to entire row if desired)
- [x] Add title (Chart title, font size, font color)

Features
- [x] Toggle row spanning
- [x] Wrap text within cells 
- [x] Resize columns (in legacy UI, the user is able to drag the column to resize within the panel. Does not need to be accomplished in the exact same way) 
- [x] Resize chart

Shubham009-prog:
## Description
Replaced normal table with Data Grid Block (MUI)

Added features:

-  Title Styling
- Header Styling
- Color By value
- Column Resizing
- Cell Styling
- Row Spanning
- Text Wrapping
- Pagination
- Resizing
- Click event
- Also added color picker as shown in figma


ehynd:
@Shubham009-prog a few changes since this pr - could you please merge in dev and retest / fix as needed please?

PR #984, opened 2025-04-24
feat(gen-code): generate code with llm option fix

### LINKED ISSUE #242: Generate Code within notebooks not working

The option to generate code from natural language in the notebooks is presently commented out, so it is not working as intended as clicking it does not do anything. 

1. Open an app in the platform
2. Click into the editing view 
3. Navigate to a notebook
4. Within a cell, type a request in natural language. 
5. Highlight the request, right click, and select "Generate Code" (or Ctrl+Shift+G)
6. Observe that nothing changes; no code is generated.

### Expected results

Upon clicking "Generate Code", code will be generated based on the highlighted text and populate below it.

### Actual results

Upon clicking "Generate Code", nothing happens. 

#### Screenshots

Image

### Requirements
- [x] Analyze best approach: either 1) generate code entirely in python OR 2) generate based on cell type. 
     - [x] If option 1: change cell type to python regardless of what the cell is and generate in python. 
     - [x] If option 2:  The LLM will produce output in whatever language the cell is.  
           -If the cell is set to Pixel -> when clicking generate code, cell type will change to Python and write Python code.  
           -If cell is set to R -> when clicking generate code, R code will be written 
           -If cell is set to Python -> when clicking generate code, Python code will be written 
- [x] Will need some sort of general prompt to call to the LLM, something like "Write me code that does {copied text from user} in {language of cell type}
- [x] After clicking generate code, the user's original text should be converted to a comment based on language.  Then, the code should populate directly below in the same cell.

Paulson-Robert:
## How to Test

2. Expected outcomes


ehynd:
hi @Paulson-Robert I get an error when trying to run this locally! did some troubleshooting and it appears to have to do with paramValues.  could you try to modify the reactor you call to generate the code to send these param values? 

ex: 
image

can hop on a call to show you my issue if needed!

Paulson-Robert:
> hi @Paulson-Robert I get an error when trying to run this locally! did some troubleshooting and it appears to have to do with paramValues. could you try to modify the reactor you call to generate the code to send these param values?
> 
> ex: image
> 
> can hop on a call to show you my issue if needed!

Updated the reactor call with paramValues and committed.

libs/renderer/src/components/cell-defaults/code-cell/CodeCell.tsx
```diff
@@ -147,41 +149,41 @@ export const CodeCell: CellComponent<CodeCellDef> = observer((props) => {
...
+            setLLMLoading(true);
+            if (!modelId) {
+                throw new Error("No Agent Model Engine");
+            }
+
+            const res = await runPixel(
+                `LLM(engine = "${modelId}", command = "${prompt}", paramValues = [ {"max_completion_tokens": 2000, "temperature": 0.3} ] );`,
+            );
+
+            const LLMResponse = res.pixelReturn[0].output["response"];
```
johbaxter, line 171: We need BE reactor to help us standardize so we do not have to do any string manipulation to grab code.  Or prefix the prompt with ONLY SEND CODE.

johbaxter (approved):

libs/renderer/src/components/cell-defaults/code-cell/CodeCell.tsx
johbaxter: @ehynd @ppatel9703 We need to follow up on this.

Parth wouldn't hurt if i could bounce this BE reactor idea off of you

johbaxter (approved):

Issue #934, opened 2025-04-21
Ability to display/parse multiple variables in a text or markdown block


kzsb03:
### Description 

Currently only one variable displays on the FE. If there are 2 it doesn't work and shows the bracket notation.

### Existing

Is there any relevant code or ui feature that the developer should know about when resolving this issue. 

### Requirements

What are more specific requirements that needs to take to consider this issue resolved.

- [ ]
- [ ]

Manikandan-Kanini:
Can variable names include spaces, such as "var one"?, Now we are able create with space.

PR #1027, opened 2025-05-02
#934 display or parse multiple variables in a text fixed

### LINKED ISSUE #1262: Update viz chart titles in block settings

### Description

Each visualization title within block settings contains the word "Echart" 

Ex: "Echart Bar Graph" "Echart Stack Chart" 

Image

We want to remove the word "Echart" from all chart titles in this block, so they just so "Bar Chart" and "Stack Chart" etc.  

Additionally, there are some blocks (Ex: Bar) where the blocks menu says "Bar Chart" but the block settings title says "Bar Graph".  We want to use the same title in both places, changing what is in block settings to match the blocks menu. 

### Relevant Feature or Code

_No response_

### Requirements (what)

- [x] Remove "echart" from title in block settings of each chart 
- [x] Make sure rest of title matches up with the title in the Blocks menu.  Ex: "Echart Bar Graph" should become "Bar Chart" to match what is in Blocks menu

Manikandan-Kanini:
## How to Test

2. Expected outcomes


johbaxter:
@Manikandan-Kanini please fix merge conflicts

Hanumant2828:
Hi @Manikandan-Kanini I have tested the above task and have observation related to task, could you please review it.
link - https://github.com/SEMOSS/semoss-ui/issues/1738
CC- @rameshpaulraj @Pranchaudhari

libs/renderer/src/store/state/state.store.ts
```diff
@@ -679,86 +679,86 @@ export class StateStore {
...
-
-					if (p.length === 0) {
-						return c.output;
-					} else {
-						const key = p[0];
-
-						if (key in c._exposed) {
-							// get the search path
-							const s = p.join(".");
+		cleaned = cleaned.replace(/{{|}}/g, "");
```
johbaxter: Not sure we should be doing this logic here.  There should be a function somewhere in code where we do this regex, not remembering off top of head.  useBlock?

Manikandan-Kanini: I removed the old logic and used flattenVariable in useBlock.

PR #1032, opened 2025-05-02
fix(client): swapping dataTestId to Id on Notifications

memisrose:
## Description
Warnings on Notifications data-testId's.

## Changes Made
Swapping data-testId to Id on the Snackbar component in the Notifications. Snackbar component does not allow the data-testId.

Adding data-testId across the client site to buttons and in the legacy app buttons regarding importing.

## How to Test

1. Save an app to trigger the Notification popup 
2. Inspect for the id
3. Inspect across the application for buttons and check for data-testIds


johbaxter:
@memisrose can you walk me through this later this week, if we can get @roxyramos on that call it would be great too, just assuming this. data-id stuff are things roxy is grabbing

johbaxter (approved):

PR #1047, opened 2025-05-06
733 input block test

Mezzet:
## Description
Added style tests for input block

## Changes Made
Added InputBlock test file

## How to Test

Create mock data in json format
Render block using custom renderer with mock data
Run tests

## Notes
Need to add onChange tests and input block size test failing due to it overflowing its parent container


johbaxter:
Whats the purpose of so many changed files?

Also why are we moving things around in the package.json.

I don't want to accept a PR for testing with additions to the package.json.  unless they are simply adding packages that are approved by Rose.

If there was an addition to the package.json lets make sure the diffs show just that. 

 If there are no additions we should not be commiting our lock file either.

Looking at other test prs they are generally isolated to just one test file added or changed not so much as this one as far as diffs

@memisrose and @Mezzet  please put this in draft until addressed

Mezzet:
Apologies, the previous push was the correct one. I will work on reverting my changes

memisrose (approved):

memisrose:
all tests passing
image

PR #1049, opened 2025-05-06
feat(renderer): add sync and async to block events

johbaxter:
## How to Test

2. Expected outcomes


johbaxter:
image

Migration function added to modify all events on blocks.  I changed the DS to track the executionType: 'sync' vs 'async'

Required me to change the types and also go through our default-menu to modify block events there.

Whats leftover:

**Modify block configs**
**Modify Template Apps**
**TEST TEST TEST**

PR #1053, opened 2025-05-06
Grid container updates

memisrose:
## Description
Updates and changes to the new grid container settings.

## Changes Made

- Updating margin and padding settings so the user can specify sides
- Adding several icons to the setting inputs to more accurately explain the input to the user
- Fixing bugs
- Removing display and align items settings from the position settings


johbaxter:
We will need a follow up ticket to iterate on this @ehynd 

- Padding setting is a little buggy
- Gap setting is a little buggy when doing anything other than percent, modify the function that determines the width for child els based on gap.
- Columns select show icon or images in menu items

PR #1087, opened 2025-05-09
793 update text field component

### LINKED ISSUE #793: Update the text field component to include "Text Field - Stacked"
### Description 

We are adding a variant or secondary option for the Text Field component that is to be specifically used when updating the drag and drop experience. This is not for regular use on other platform pages - the standard Text Field component should be used on all other pages.

This new option allows us to stack the label overtop the text input streamlining the legibility and usability of that component within the small sidebar.

Ex: Image

At the request of @neelneelneel we also made a variant where the Text Field - Label Above Input is preconfigured with the button group which is also to be used exclusively in the drag and drop experience.

Ex: 

Image

### Existing

Figma: [SEMOSS Design System with Usage Guidance](https://www.figma.com/design/kZwcxDBSMJbOcFaCin2xbd/SEMOSS-MUI-Core-v5.4.0?m=auto&node-id=6570-46740&t=oe7zw4z212K41lTz-1)

bannaarisamy-shanmugham-kanini:
## Description
Creating a new textfield component to have label above the text field and with few other design changes

## Changes Made
Created a new Textfield named TextFieldStack which will have a label above the textfield

## How to Test

Since this is for dev's development purpose, this will be closing used by dev while creating new settings section


libs/ui/src/components/TextFieldStack/TextFieldStack.tsx
```diff
@@ -0,0 +1,78 @@
...
+    SxProps,
+    InputLabel,
+    Typography,
+} from "@mui/material";
+import { InfoOutlined } from "@mui/icons-material";
+import styled from "@emotion/styled";
+
+export type TextFieldStackProps = MuiTextFieldProps & {
+    /** custom style object */
+    sx?: SxProps;
```
johbaxter, line 15: We should be adding a optional param for hint, to conditionally show the tooltip

libs/ui/src/components/TextFieldStack/TextFieldStack.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+const StyledInputLabel = styled(InputLabel)(({}) => ({
+    display: "flex",
+    flexDirection: "row",
+    gap: "4px",
+    marginBottom: "8px",
+}));
+
+const StyledTypography = styled(Typography)(({}) => ({
+    color: "#666666",
```
johbaxter: Can we pull from our theme.ts

libs/ui/src/components/TextFieldStack/TextFieldStack.tsx
```diff
@@ -0,0 +1,78 @@
...
+        }
+    }, [componentId]);
+
+    return (
+        <>
+            <StyledInputLabel shrink={false} htmlFor={componentId}>
+                <StyledTypography variant="body2">
+                    {props.label ?? "Label"}
+                </StyledTypography>
+                <InfoOutlined fontSize="small" color="action" />
```
johbaxter: Utilize <Tooltip /> component based on added hint prop

Keep the Icon you use here but wrap it in <Tooltip /> ^

libs/ui/src/components/TextFieldStack/TextFieldStack.tsx
bannaarisamy-shanmugham-kanini: I have resolved this issue

libs/ui/src/components/TextFieldStack/TextFieldStack.tsx
bannaarisamy-shanmugham-kanini: I have resolved this issue

libs/ui/src/components/TextFieldStack/TextFieldStack.tsx
bannaarisamy-shanmugham-kanini: I have resolved this issue

libs/renderer/src/components/block-defaults/input-block/InputBlock.tsx
johbaxter, line 1: What are the reasons for the changes in this file?

bannaarisamy-shanmugham-kanini: I have made changes for doing few tweaks towards UI, since I considered Textfieldstack as a block and not as a component. Since Textfieldstack is a component, I reverted Inputblock to old code.

johbaxter:
We now have this as a component in our Library, we will need a dev ticket to update all spots that should be using this new component in our BlockSettings.

- [ ] Update Block Settings to utilize TextFieldStack

PR #1120, opened 2025-05-13
997 show console in notebook as it is running

### LINKED ISSUE #997: Show console in notebook as it is running
### Description 

If I am running a long piece of python code, I want it to show me the steps as it goes.  

Ex: If I have 

*code snippet1*
print (test)
*codesnippet2*
print (test2)

Once code snippet1 finishes, I want to print test while codesnippet2 is still running. Then when that finishes, test2 will show up.

This exists in the console, the user just does not see it.  Instead, the entire cell would run and then it would print test and test2 at the same time. 

## Sample Code that can be used when testing 

# First block of code
def calculate_sum(n):
    total = 0
    for i in range(1, n + 1):
        total += i
    return total

def find_max(numbers):
    max_num = numbers[0]
    for num in numbers:
        if num > max_num:
            max_num = num
    return max_num

def reverse_string(s):
    return s[::-1]

# Example usage of the above functions
sum_result = calculate_sum(10)
max_result = find_max([3, 5, 7, 2, 8])
reversed_string = reverse_string("hello")

#this should print below the notebook cell when the above finishes running 
print(f"Sum: {sum_result}, Max: {max_result}, Reversed: {reversed_string}")

# Second block of code
def factorial(n):
    if n == 0:
        return 1
    else:
        return n * factorial(n - 1)

def is_prime(num):
    if num <= 1:
        return False
    for i in range(2, int(num ** 0.5) + 1):
        if num % i == 0:
            return False
    return True

def fibonacci(n):
    fib_sequence = [0, 1]
    while len(fib_sequence) < n:
        fib_sequence.append(fib_sequence[-1] + fib_sequence[-2])
    return fib_sequence

# Example usage of the above functions
factorial_result = factorial(5)
prime_check = is_prime(11)
fibonacci_sequence = fibonacci(10)

#this should then print below the first print statement when it finishes running 
print(f"Factorial: {factorial_result}, Is Prime: {prime_check}, Fibonacci: {fibonacci_sequence}")

JaganSomannaKanini:
## Description
When we have multiple print statements in a cell, now we can see the output one after the other.

## Changes Made
Added a condition for the code cell to display the output one after the other.

## How to Test
The code given in the ticket is small to display the output one after the other, so we have to use a delay to see the changes in the output.

## Notes
To add delay use "import time" in the first line and use "time.sleep(1)" in between the print statements. This will create 1 second delay.


ehynd:
@johbaxter put this in a notebook cell to test. there will be a ~5 second lag between the first print statement showing up and the second (due to the time.sleep(5) added in for testing)  

import time
def calculate_sum(n):
    total = 0
    for i in range(1, n + 1):
        total += i
    return total

def find_max(numbers):
    max_num = numbers[0]
    for num in numbers:
        if num > max_num:
            max_num = num
    return max_num

def reverse_string(s):
    return s[::-1]

sum_result = calculate_sum(10)
max_result = find_max([3, 5, 7, 2, 8])
reversed_string = reverse_string("hello")

#this should print below the notebook cell when the above finishes running
print(f"Sum: {sum_result}, Max: {max_result}, Reversed: {reversed_string}")
time.sleep(5)


def factorial(n):
    if n == 0:
        return 1
    else:
        return n * factorial(n - 1)

def is_prime(num):
    if num <= 1:
        return False
    for i in range(2, int(num ** 0.5) + 1):
        if num % i == 0:
            return False
        return True

def fibonacci(n):
    fib_sequence = [0, 1]
    while len(fib_sequence) < n:
        fib_sequence.append(fib_sequence[-1] + fib_sequence[-2])
    return fib_sequence

import time
factorial_result = factorial(5)
prime_check = is_prime(11)
fibonacci_sequence = fibonacci(10)

#this should then print below the first print statement when it finishes running
print(f"Factorial: {factorial_result}, Is Prime: {prime_check}, Fibonacci: {fibonacci_sequence}")

ehynd:
@johbaxter ping me when you do this one and i have test code for you but it is pasting wrong here

packages/client/src/components/notebook/NotebookCell.tsx
```diff
@@ -780,52 +780,115 @@ export const NotebookCell = observer(
                             </StyledRunIconButton>
                             <StyledCardInput>{rendered}</StyledCardInput>
                         </StyledCardContent>
-                        {cell.isExecuted && (
+                        {cell.widget === 'code' ? (
```
johbaxter, line 783: Think about other widgets other than notebook cells.  We need to be able to pull the logs for things other than python code cells

johbaxter (changes requested):
Please make sure this works for all notebook widgets. not just code..  Take a look at my comment in the code

JaganSomannaKanini:
Sure @johbaxter , will take a look on it.

johbaxter (approved):
Curious about why we have to have the outer condition if cell.widget === 'code'

image

JaganSomannaKanini:
> Curious about why we have to have the outer condition if cell.widget === 'code'
> 
> image

In the ticket it was mentioned for a long Python code @johbaxter, so the condition made sense in that manner. Without that condition it would have affected the ability to show the output for other widget cells because in some cases the output would come through console and in other case the complete output would come only after the cell is executed.

PR #1122, opened 2025-05-13
975 update the select component

### LINKED ISSUE #975: Update the Select component to include "Select - Stacked"
## Description
We are adding a variant or secondary option for the Select component that is to be specifically used when updating the drag and drop experience. This is not for regular use on other platform pages - the standard Select component should be used on all other pages.

This new option allows us to stack the label overtop the select input streamlining the legibility and usability of that component within the small sidebar.

Ex:

Image

Do confirm if this already exists in our current component library and ensure it is set up per the design specs in Figma.

## Existing
Figma: https://github.com/SEMOSS/semoss-ui/issues/842

bannaarisamy-shanmugham-kanini:
## Description
Create a new select field component with label and select field in Column way instead of outlined design

## Changes Made
Created a new SelectStack field component in libs/ui folder with needed changes

## How to Test

1. Developer can access the changes from semoss/ui library
2. Developer can import SelectStack from semoss/ui package.


johbaxter (approved):

johbaxter:
Ticket needed to update block settings to use this component

PR #1135, opened 2025-05-14
feat(client): lazyloading for modelsettings

### LINKED ISSUE #116: [TASK] Lazy Loading is a bit deceptive
Lazy loading is a bit deceptive, it can appear you are at the end of the list if you haven't scrolled completely to the bottom and there's not an easy way of knowing how many there are in total.

Is there a way to show that more are loading and/or show a full number count so user has indication when they just need to keep scrolling to pull more?

SuryaNarayanan-Kgs:
## How to Test

2. Expected outcomes


packages/client/src/pages/settings/SkeletonCard.tsx
```diff
@@ -0,0 +1,84 @@
+import { Box, Skeleton, Card } from '@semoss/ui';
```
AAfghahi, line 1: Is this a reusable component? Are other features going to reasonably want this? If so, I think that we should put it in semoss/ui as opposed to making it a page in settings.

AAfghahi (commented):
Could you show me some steps to reproduce this, and also include some screenshots of what it looks like. 

Thank you!

johbaxter:
@SuryaNarayanan-Kgs

Hey Surya i refactored the code that you had done. Please pay attention to how i cleaned the code up.

Please no inline styles.

And when we should be making a reusable component, please place it in `libs/ui` as Arash stated.

I am also noticing another thing: 

The load state doesn't seem to stop if you have gotten all your results

image

@AAfghahi We likely need to standardize to this component on the AppCatalog Page as well, youll see this code pushed up soon with new landing page.

johbaxter (changes requested):
Please delete this file and look at the additions that i made to component library.

Also address the bug that is with always showing load Skeleton Cards. Play with the limit in `EngineSettingsIndexPage`

PR #1136, opened 2025-05-14
1064 scatter plot render bug

### LINKED ISSUE #1064: Scatter Plot Render Bug

1. Create dataframe with diabetes data
2. Drag in scatter plot 
3. Drag height and weight to x and y fields 

### Expected results

A scatter plot is rendered. 

### Actual results

Chart stays blank
Error in console is as following: 

[ECharts] `weight` is invalid id or name. Must be a string or number.
[ECharts] `height` is invalid id or name. Must be a string or number.

### Environment

This is seen on dev branch. 

### Requirements

- [x] Resolve bug so scatter plot can render as expected

Shubham009-prog:
## Description
Scatter plot not rendering

## Changes Made
Added some data check as there was some change in data due to which plot was not rendering.

## How to Test
Drag and drop scatter plot it will start rendering


Shubham009-prog:
Hi, I have raised the PR of this Ticket. Also checked that the Brush event is working fine.

johbaxter (approved):

PR #1142, opened 2025-05-14
fix(client): make the edit state of the SMSS under an engine more clear

Gunasrini18:
## How to Test

2. Expected outcomes


johbaxter (approved):
Looks good

PR #1151, opened 2025-05-15
Ability to select multiple blocks on Drag and Drop App Builder

### LINKED ISSUE #1054: [Enhancement] Keyboard shortcut to delete blocks
### Description 

Enhance the user experience by providing a convenient keyboard shortcut for deleting one or multiple selected blocks.
 
### Existing

Currently, the shortcut Ctrl + Alt + D is used for JSON, so Ctrl + D is not preferred for delete due to its similarity. A new shortcut needs to be identified and implemented to avoid conflicts and ensure intuitive usability.

Suggested: Ctrl + X since it is the the shortcut for "Cut" which is similar to delete. 

### Requirements

- [x] Identify and implement a new keyboard shortcut that does not conflict with existing shortcuts.
- [x] Ensure the shortcut allows for the deletion of both single and multiple selected blocks.

NaveenRamasamy028:
Now the user can able to select multiple blocks.


johbaxter (approved):
Nice feature @NaveenRamasamy028

PR #1205, opened 2025-05-22
feat(client): User permission set correctly also if user is not present on the 1st page of pagination in app/engine settings

### LINKED ISSUE #1201: Permissions will not be set correctly if the user is not present on the 1st page of pagination in app/engine settings

According to the logic written in MembersTable.tsx -> setUserDetails, we are currently checking if the logged in user is part of the MembersTable. The members fetched are only 5 at a time as the table has pagination. Hence it can be the case that the logged in has a different permission but since there are only 5 visible, hence the logged in user was given a wrong permission.

1. Go to any app
2. Go to the settings of app using edit icon.
3. Notice if your name is not in the first list, your permission will not be set correctly

### Expected results

Expected result of action.

### Actual results

What actually happened.

#### Screenshots

Image

### Environment

- browser type and version:
- node.js version: `node -v`

### Checklist

### Requirements

- [ ]
- [ ]

pallabi2303:
## Description
The logged-in user's permission is determined by a dedicated API call, ensuring correct access rights are set even if the user does not appear on the first page of the members list. This guarantees accurate UI controls regardless of pagination.

## Changes Made
1. Added a dedicated API call to fetch the logged-in user's member record directly, using their user ID and a limit of 1.

2. Implemented the setUserDetails function to set the logged-in user's data and permission based on the result of this 
    dedicated API call, ensuring correct permission assignment even if the user is not present in the paginated members list.

## How to Test
1. Log in as a user with elevated permissions (Author or Editor).
2. Navigate to the Members settings page for an app or engine.
3. Ensure the members list is long enough that the logged-in user does not appear on the first page
4. User can see that they can access member setting permission even if user doesn't appear on the first page of members list


packages/client/src/components/settings/MembersTable.tsx
```diff
@@ -295,65 +295,69 @@ export const MembersTable = (props: MembersTableProps) => {
...
-        // select the member when done mounting
-        memberSearchRef.current?.focus();
-    }, [getMembers.status, getMembers.data]);
-
-    // useLayoutEffect(() => {
-    //     if (getMembers.status !== 'SUCCESS' || !getMembers.data) {
-    //         return;
-    //     }
+        setUserDetails();
+    }, [userDetails.status, userDetails.data]);
```
anurag91jain: Let's try to remove the dependencies and check it once. We don't need to call setUserDetails again and again on pagination as it wouldn't change

anurag91jain (changes requested):

packages/client/src/components/settings/MembersTable.tsx
pallabi2303: Yes , I have removed userDetails.data dependency as it wouldn't affect the behavior, so that  setUserDetails wouldn't be called again and again.

packages/client/src/components/settings/MembersTable.tsx
```diff
@@ -295,65 +295,69 @@ export const MembersTable = (props: MembersTableProps) => {
         setPage(0);
     }, [debouncedSearch]);
 
+    const getUserDataApi: Parameters<typeof useAPI>[0] =
```
AAfghahi: This double ternary seems worse than just a if else, and it is harder to read.

pallabi2303: Now I've changed the double ternary to simple if-else statements.

packages/client/src/components/settings/MembersTable.tsx
```diff
@@ -272,88 +309,59 @@ export const MembersTable = (props: MembersTableProps) => {
...
-                  rowsPerPage, // limit
+                  0, // offset
+                  1, // limit
               ]
             : type === 'APP'
             ? [
                   'getProjectUsers',
                   adminMode,
                   id,
-                  debouncedSearch ? debouncedSearch : undefined,
```
AAfghahi, line 285: why did we get rid of this?

anurag91jain: Fixed it @AAfghahi and consolidated 2 if conditions into 1.

anurag91jain (approved):

johbaxter:
May be a separate issue.

Im logged in as cc:
image

Should i be able to demote authors, or should i only be able to demote users of the same status as me along with myself:
image

I can push this, but yeah still think there's issues.

Lmk what you think @AAfghahi

pallabi2303:
Yes, the issue has been resolved. Editors no longer have the ability to demote Authors.

johbaxter:
Looks good ty

PR #1208, opened 2025-05-23
feat(client): after adding member, admin doesn't turn off

### LINKED ISSUE #1203: After adding user, admin turns off
When I add another user as an Admin, the page refreshes which causes the Admin to turn off and hence throw an error

1. Go to settings
2. Turn on Admin mode and open an app which you don't have access to
3. Add a member other than yourself
4. Error: The page refreshes and Admin is off and shows an error "you don't have access"

### Expected results

Page should not refresh or admin should not be turned off

### Actual results

Page refreshes and admin is turned off

#### Screenshots

Image

This logic is causing the page to refresh. The store (which contains whether admin is on or off) doesn't persist in localStorage or redux-persist.

### Environment

- browser type and version:
- node.js version: `node -v`

### Checklist

### Requirements

- [ ]
- [ ]

pallabi2303:
## Description
feat(client): after adding member, admin doesn't turn off

## Changes Made
After Adding user, the page doesn't get refreshed automatically. So, 'Admin On' remains as it is.

## How to Test
1. Log in as an Admin.
2. Navigate to the Members settings page for an app or engine that user don't have an access for.
3. On 'Admin On', Add Member.
4. User can see that after Member being added Admin remains on.

### DIFF
--- a/packages/client/src/components/settings/MembersAddOverlay.tsx
+++ b/packages/client/src/components/settings/MembersAddOverlay.tsx
@@ -469,10 +469,6 @@ export const MembersAddOverlay = (props: MembersAddOverlayProps) => {
         } finally {
             // close the overlay
             closeOverlay(type, success);
-            //adding a refresh to the page so that the updated by and timestamp show up
-            if (success) {
-                window.location.reload();
-            }
         }
     };
 


pallabi2303:
@ishumita 
You mentioned you have added the refresh so that the updated by and timestamp show up, but even if we don't refresh the page it is already behaving like that. So, I have removed the refresh as the Admin was turning off because of it, and we were getting error.

ishumita:
@pallabi2303  sure, thank you for letting me know:)

anurag91jain (approved):

AAfghahi (approved):

johbaxter (approved):

PR #1230, opened 2025-05-28
BugFix for PROD catalog

ammb-123:
## Description

DTTL Approved catalog items
## Changes Made

DT AI Core admins should be able to access AI Core services as outlined in the attached document.
Only DTTL approved resources, as specified in the attached sheet will be visible in DT AI Core and granted access.
Users should not be able to see any catalog items.

## How to Test

DTTL approved resources, as specified in the attached sheet will be visible in DT AI Core and granted access.
Users should not be able to see any catalog items.

## Notes

[Uploading NonApproved_Doc1.docx…]()

### DIFF
--- a/packages/client/src/pages/import/ImportPage.tsx
+++ b/packages/client/src/pages/import/ImportPage.tsx
@@ -343,12 +343,22 @@ export const ImportPage = () => {
         }
         //Function
 
-        if (
-            obj?.FUNCTION?.Function
-        ) {
+        if (obj.FUNCTION.Function.length > 0) {
             for (let k = 0; k <= obj.FUNCTION.Function.length; k++) {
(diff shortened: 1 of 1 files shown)


johbaxter:
Hello can i get this list? of Non approved LLMs, it does not actually let me download the file

We likely also want this on the Backend somewhere on RDF_MAP

johbaxter:
@ppatel9703 it looks like we are limiting access to some of the llms in our catalog.

I know our import catalog as it stands now is on the FE as a constant.

We need to move that list of catalog items to the backend at some point.

But to not break anything.

Should we add to the RDF_MAP or add some constant somewhere of approved resources

johbaxter (changes requested):
Def think there needs to be backend additions to this to account for the approved resources.

Ideally would need to pass the approved resources from backend to frontend

anurag91jain:
This is not required

PR #1238, opened 2025-05-30
697 aggregate column function

### LINKED ISSUE #697: Aggregate Column Functions
### Description 

We need a way for low/no code users to be able to select what function they are wanting when selecting data fields within the frame. This will allow them to utilize one frame for multiple visualizations and also not require them to code within the notebook. 

### Existing

This functionality exists within legacy system, please start there to see how it is working or ask Elise for a demo.  

### Requirements

Once the user selects a frame and a column, they should be able to select what function they want done on the column. 

If the column is a number column, it will have different functions than a string or date column. 

The default function should be average for number columns and count for string/date columns. 

Number columns necessary functions:
- [x] Average 
- [x] Sum
- [x] Count
- [x] Unique Count
- [x] Minimum
- [x] Maximum
- [x] Median

String/date columns necessary functions:
- [x]  Count
- [x] Unique Count

Paulson-Robert:
## How to Test

2. Expected outcomes


johbaxter (approved):

johbaxter:
Thank you Paulson, you the man!

PR #1257, opened 2025-06-04
Linting updates

### LINKED ISSUE #762: Unified lint rules across code base
We need commits to be linted only on staged files

While having this we need all packages/libs to go off of one unified lint ruleset

For `packages/client` we have a custom rule that needs to be applied:

@mui/material or @mui/material/whatever-module should not be allowed to be imported into that specific portion of codebase

https://eslint.org/docs/latest/rules/no-restricted-imports

memisrose:
## Description

- The goal is to lint only staged files throughout the entire app
- Several of the projects do not have lint targets; these files should be ignored from linting
- Do not block a user's commit unless the issue touches a file they have worked on

## Changes Made

- Adding an eslintrc.json at the root: this is needed for lint-staged but does not have to be specific or detailed, the eslintrc.json at the project level will override
- npx lint-staged --concurrent false --relative
--lint-stage: grab the list of staged files
--concurrent false : override lint-staged default and run linting sequentially
--relative: the default flag to pass relative file paths

- lintstagedrc
--rules for lint-staged
-- {*,libs,packages} : define the directories
--`nx affected` grab only files that have changed
-- -target=typecheck -->remove?
--running `npx nx affected --target=typecheck` will show which projects are set up (currently renderer and client)

- consolidating eslint
--created an eslintrc.json at the root of the project
--to test this, run 'pnpm run lint' for each project to ensure it is still working

## How to Test

1. Add changes 
2. Commit changes


AAfghahi (approved):

johbaxter:
@memisrose and @AAfghahi i left this out and my apologies.  We need to make sure mui/material does not get committed in `packages/client`

image
Something like this i attempted in the root eslintrc, but i was able to commit a mui/material import.

If you look at our eslintrc in client we had this which did work. Look below

image

johbaxter (changes requested):
Take a look at comment in PR.  We need the restricted import for client 
https://github.com/SEMOSS/semoss-ui/blob/dev/packages/client/.eslintrc.json

RNubla:
Reviewed and tested and it looks good. Just in case, for **mac**/**linux** machines, if you see this message when you commit
image
Please run the following command in the root directory
```
chmod ug+x .husky/*
```
Source: https://github.com/typicode/husky/issues/1177#issuecomment-1212831091

RNubla:
@johbaxter created a PR for 783 branch

PR #1264, opened 2025-06-05
Fix resolve ask CSV app template not working

Vishal24-Kanini:
## Description
Fix code in AskCSVTemplate.tsx , which ensures "Ask" button should not stuck now when clicks also the second line of code in the notebook should not stuck in loading as well

## Changes Made
- I have I modified the limits of MAX_INPUT_TOKENS and MAX_TOKENS in the model SMSS Properties
- I ran that modified model and getting the expected output
- then copy that JSON from FE and replace it down in the AskCSVTemplate.tsx


johbaxter:
Took some of this code in previous weeks but the app still does not work.  We will ticket again in future

PR #1273, opened 2025-06-06
feat(client): Member settings restriction on deleting all admins from Members Table

### LINKED ISSUE #1156: Member Settings Restriction on Deleting Selected
In the Member Settings page, there is an option to Delete Selected members from the table:

Image

This will delete every member **except** the remaining admin member. A pop-up should appear indicating to the user that they cannot delete the last admin in the table.

pallabi2303:
## Description

Restricted The deletion of all admins from the Members Table

## Changes Made

1. Added a new API configuration, getAllAuthorsApi, to fetch all users with the "OWNER" (Author) permission. This is used to determine the current list of authors in the system.
2. In the openDeleteMembersModal function, before allowing deletion, the code checks if the selected members to delete include all current authors. If so, it prevents the action and shows an error notification: ' "You cannot delete all the admins(Authors) from the table."

## How to Test

1. Log in as an Author
2. Go to settings and then App Settings
3. Try to delete all admins(Authors)
4. It will show a pop-up saying 'You cannot delete the all admin (Author) in the table.'

image


anurag91jain (approved):

packages/client/src/components/settings/MembersTable.tsx
```diff
@@ -308,10 +318,34 @@ export const MembersTable = (props: MembersTableProps) => {
...
+    );
+
+    useEffect(() => {
+        if (
+            allAuthorsResponse.status === 'SUCCESS' &&
+            allAuthorsResponse.data
+        ) {
+            setAllAuthors(allAuthorsResponse.data['members']);
+        } else {
+            setAllAuthors([]);
```
AAfghahi, line 329: is there a reason that we are setting this as an empty array as opposed to just using null as it is set to? That way you can just check to see if it exists, rather than checkign the length constantly.

pallabi2303: Yes, there is a reason.
Setting allAuthors as an empty array [] makes it easy to add the restriction on the authors length as we can directly use length for the array, but if we use null, then we will have to check for null every time before using the function to avoid errors, making it more complex.

AAfghahi: But if that is the reason, then why do we set it to null as default?

pallabi2303: We have set null to the getAllAuthorsApi, not to the allAuthors, allAuthors is initialized as an array.

AAfghahi (commented):
image


It let me delete all users

PR #1275, opened 2025-06-06
feat(client): discoverable apps notification fixed

### LINKED ISSUE #644: [BUG] Issues with 'Discoverable apps'

1. Click on ‘Discoverable’ apps 
2. 

### Expected results

Users should be able to request for access

### Actual results

System throws an error: "Project does not exist or user does not have access to the project"

#### Screenshots

Error while requesting access
Image

### Environment

- browser type and version:
- node.js version: `node -v`

### Tasks

- [ ]
- [ ]

shubham-sp-07:
## Description

The user does not have access to view this project or project id is invalid  notification msg fixed for discoverable page

## How to Test

2. Expected outcomes


anurag91jain (approved):

shubham-sp-07:
@CodiumAI-Agent /improve

shubham-sp-07:
Please Refer first commit for issue fix

packages/client/src/pages/app/AppDetailPage.tsx
```diff
@@ -239,105 +239,99 @@ export const AppDetailPage = () => {
...
+        ];
+        if (permission !== 'discoverable') {
+            promises.push(fetchDependencies(monolithStore, id));
+        }
+        const results = await Promise.allSettled(promises);
+        results.forEach((res, idx) => {
+            if (res.status === 'rejected') {
+                emitMessage(true, res.reason);
+            } else {
+                if (idx === 0) {
```
AAfghahi, line 260: is the array always the same way? Its never structured differently? One thing that worries me about this if else statement is that if the structure ever changes it would break. 

Is there another check that we can make that is a bit more stringent?

packages/client/src/pages/app/AppDetailPage.tsx
```diff
@@ -239,105 +239,99 @@ export const AppDetailPage = () => {
...
+                    } else {
+                        const modelled = modelDependencies(res.value.output);
+                        setValue('dependencies', modelled);
+                        setValue('selectedDependencies', modelled);
+                    }
                 }
-            }),
-        );
+            }
+        });
```
AAfghahi, line 334: Could we add error checking to this as well? Does not look like we do anything if the pixel call errors out.

AAfghahi (changes requested):
see comments

packages/client/src/pages/app/AppDetailPage.tsx
anurag91jain: @AAfghahi the indexes are based on the array insertion in the promises. It will always remain the same till the array is changed

packages/client/src/pages/app/AppDetailPage.tsx
anurag91jain: Individual calls are tracked within the indexes and if the overall call fails, that is taken as rejected.

AAfghahi (approved):

PR #1289, opened 2025-06-10
fix: fixing issue with getting user data

### LINKED ISSUE #1201: Permissions will not be set correctly if the user is not present on the 1st page of pagination in app/engine settings

neelneelneel:
## Description
Fix with pulling members

### DIFF
--- a/packages/client/src/components/settings/MembersTable.tsx
+++ b/packages/client/src/components/settings/MembersTable.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useMemo, useState, useRef, useLayoutEffect } from 'react';
+import { useEffect, useMemo, useState, useRef } from 'react';
 import { Add, Delete, Edit } from '@mui/icons-material';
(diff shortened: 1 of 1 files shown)


johbaxter:
image

In a PR where we do code cleanup, please use correct api to get user data instead of indexing

PR #1290, opened 2025-06-11
feat(client): delete_multiple_job_changes

### LINKED ISSUE #1143: Create multi-delete functionality for jobs (FE)
### Description 

This is the front-end portion of this overall ticket: https://github.com/SEMOSS/community/issues/356#issue-3053053251

The back-end portion necessary for the reactor call is here: https://github.com/SEMOSS/Semoss/issues/817

### Requirements

- [ ] Create ability to select every job (might already exist)
- [ ] If multiple jobs selected, the delete jobs verbiage to reflect "jobs" instead of "job"
- [ ] If needed, amend delete reactor call to take in an array of jobs to accommodate multiple jobs

aditiiisinhaa:
## Description
If two or more jobs are getting deleted successfully then a notification appears saying 'Successfully deleted all selected jobs' and if some jobs fail to get deleted then the notification appears as 'Some jobs were deleted successfully, but the following jobs could not be deleted: (failedJobNames)'

## How to Test
1. Go to Open Settings then select Jobs.
2. Add two or more Jobs.
3. Select two or more Jobs.
4. Click on Delete Selected.
5. Notification will pop up.

### DIFF
--- a/packages/client/src/pages/jobs/JobsPage.tsx
+++ b/packages/client/src/pages/jobs/JobsPage.tsx
@@ -168,14 +168,27 @@ export function JobsPage() {
             .then((response) => {
                 const type = response.pixelReturn[0].operationType;
                 const output = response.pixelReturn[0].output;
-                if (type.indexOf('ERROR') === -1 && output === true) {
-                    notification.add({
-                        color: 'success',
-                        message:
-                            jobId.length > 1 && jobGroup.length > 1
-                                ? `Successfully deleted all selected jobs`
-                                : `Successfully deleted ${type}`,
-                    });
+                // Expecting output to have { success: string[], failed: string[] }
+                const successIds = output?.success || [];
+                const failedIds = output?.failed || [];
+
+                if (type.indexOf('ERROR') === -1) {
+                    if (failedIds.length === 0) {
+                        notification.add({
+                            color: 'success',
+                            message: `Successfully deleted all selected jobs`,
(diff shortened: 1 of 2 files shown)


packages/client/src/pages/jobs/JobsPage.tsx
```diff
@@ -168,14 +168,27 @@ export function JobsPage() {
...
+                        const failedJobNames = jobs
+                            .filter((job) => failedIds.includes(job.id))
+                            .map((job) => job.name)
+                            .join(', ');
+                        notification.add({
+                            color: 'warning',
+                            message: `Some jobs were deleted successfully, but the following jobs could not be deleted: ${failedJobNames}`,
+                        });
+                    }
                     jobId.length > 1 && jobGroup.length > 1
```
anurag91jain, line 192: @SindhuHarika @aditiiisinhaa can we see if these lines are required?

anurag91jain (changes requested):

packages/client/src/pages/jobs/JobsPage.tsx
themaherkhalil: Since it is multiple jobs in one request, the BE PR associated with this task (https://github.com/SEMOSS/Semoss/pull/823) returns a map with which are successful or failed. The above lines would be required in order to ensure we inform the user which ones did fail

PR #1291, opened 2025-06-11
Autocomplete notebook

### LINKED ISSUE #703: Enable pixel to auto-complete reactor methods (like on FE)
### Description 

There is no help in auto completing reactor methods.  

Pixel side of notebook doesn't help you auto-complete the reactor methods/functions: 

Image

Terminal within apps does not do this either. 

### Existing

We do have basic autocompletion working in our legacy terminal app, so this is a place that we could look into when creating this ask.

This is the old BI:  

Image

### Requirements

- [ ] Add pixel auto complete to pixel notebook cells
- [ ] Add pixel auto complete to terminal

DanielEap:
## Description
Allows autocompletion of functions/reactors in code editor for notebook

## Changes Made
Call HelpJson reactor to get all functions, changed logic of autocomplete

## How to Test
Go to Notebook, choose a language. Type something and see a list of autocomplete

## Notes
Using runPixelTwo to get all the functions, instead of configStore because configStore was inaccessible
There is styling issue, the autocomplete list only shows like 2 1/2 autocomplete options

https://github.com/user-attachments/assets/52c3acef-e745-488e-b091-3678d8de7460


AAfghahi (commented):
This looks very cool. Couple of things I noticed in testing: 


I think that the drop down is a z-index issue, it seems to go behind the larger page. Maybe that is osmething that we can adjust to make it work better. 


I noticed that when I made a new cell in a notebook, that it wasn't loading any of the functions.

johbaxter:
> This looks very cool. Couple of things I noticed in testing:
> 
> I think that the drop down is a z-index issue, it seems to go behind the larger page. Maybe that is osmething that we can adjust to make it work better.
> 
> I noticed that when I made a new cell in a notebook, that it wasn't loading any of the functions.

1. Yeah can we make sure this gets fixed, i want to be able to see all the options.  zIndex on that menu as Arash said may be the fix.

2. I also have not tested it yet, but will need to be able to have the variable suggestions in that menu still {{variable-name}}

3.  This will also require some work from backend likely.  But it would be possible to autocomplete the reactor with the specific keys its looking for

For example instead of just populating --> `MyEngines()`

We populate --> `MyEngines(type=[], limit=[], ...otherParams=[])`

Manikandan-Kanini:
@j-adam-connor @johbaxter Design issues resolved, please review the changes.

PR #1323, opened 2025-06-19
feat(client/renderer): added open event to the listeners to perform o…

### LINKED ISSUE #1298: Create onClick action for open page

### Description

Right now, users have the option to run queries or cells upon a button click.  

We want to add another option: "Open" 

This will allow the user to direct to either a URL or another page within their app.  

Ex: I want to create a form. Once my user clicks submit on the form, I want a new page to appear that says "thank you for your submission!" 

### Relevant Feature or Code

Current onClick options: 

Image

### Requirements (what)

- [x] Add a new option for "Open" to the onClick event 
- [x] Once this is selected, a dropdown should appear for "Destination"  Within Destination, there should be two options: Link or App Page 
- [x] If Link is selected, a third box appears called "URL" where the user can paste a URL
- [x] If "App Page" is selected, the third box appears showing a dropdown of all pages within the app.

Paulson-Robert:
…pen action

## How to Test

2. Expected outcomes

### DIFF
--- a/libs/renderer/src/components/block-settings/ListenerActionOverlay.tsx
+++ b/libs/renderer/src/components/block-settings/ListenerActionOverlay.tsx
@@ -64,7 +64,7 @@ export const ListenerActionOverlay = observer(
         const { state } = useBlocks();
         const { listeners, setListener } = useBlockSettings(id);
 
-        const destinationTypes = ["External", "App Page"];
+        const destinationTypes = ["External", "Internal"];
         // get the queries as an array
         const queries = computed(() => {
             return Object.values(state.queries).sort((a, b) => {
@@ -88,7 +88,6 @@ export const ListenerActionOverlay = observer(
         // Each listener have its own useForm
         const lis = listeners[listener].order[actionIdx];
 
-        console.log(lis);
         // create a new form
         const { control, handleSubmit, reset, watch, setValue } =
             useForm<ListenerActionForm>(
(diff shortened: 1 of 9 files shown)


libs/renderer/src/components/block-settings/ListenerActionOverlay.tsx
```diff
@@ -102,6 +103,16 @@ export const ListenerActionOverlay = observer(
                                   },
```
RNubla, line 102: can this whole section be refactored for better readability like
```ts
const { control, handleSubmit, reset, watch, setValue } = useForm<ListenerActionForm>({
            defaultValues: (() => {
                if (lis) {
                    if (lis.message === ActionMessages.RUN_CELL) {
                        return {
                            message: ActionMessages.RUN_CELL,
                            payload: { queryId: "", cellId: "" },
                        };
                    }if (
                        lis.message === ActionMessages.DISPATCH_OPEN_EVENT
                    ) {
                        return {
                            message: ActionMessages.DISPATCH_OPEN_EVENT,
                            payload: { destinationType: "", destination: "" },
                        };
                    }
                }
                return {
                    message: ActionMessages.RUN_QUERY,
                    payload: { queryId: "" },
                };
            })(),
        });
```

libs/renderer/src/components/block-settings/ListenerActionOverlay.tsx
```diff
@@ -122,6 +133,11 @@ export const ListenerActionOverlay = observer(
 
         // the type
         const message = watch("message");
+        const distinationType = watch("payload.destinationType");
```
RNubla, line 135: please be mindful when using `watch()`, this will re-render at the root level, which can cause performance issues. As a rule of thumb, use the `useWatch` hook as much possible

libs/renderer/src/components/block-settings/ListenerSettings.tsx
```diff
@@ -342,8 +349,22 @@ export const ListenerSettings = observer(
...
+                                                        sx={{
+                                                            display: "block",
+                                                            flex: "unset",
+                                                            width: "80%",
+                                                        }}
                                                     >
-                                                        {content}
+                                                        {isLink(content) ? (
+                                                            <Link
+                                                                target="_blank"
```
RNubla, line 361: add `rel="noopener noreferrer"` attribute to `<Link/>`

RNubla (commented):
Reviewed and tested changes. Some minor changes suggestion should be made for better readability.

PR #1341, opened 2025-06-23
feat(themefile): theme file adding primaryblack custom palette

### LINKED ISSUE #1139: Add primary-black as a custom palette to theme file
## Problem (Why)

With the redesign of landing page and refresh of UI in general, we are introducing a black button. As noted in the design ticket, we don't want to lose the capabilities of the primary and secondary colors as these are used throughout our application still so we have opted to specifically ADD a primary-black to be used for the button only (instead of replacing a current primary or secondary palette which will cause a lot of conflicts).

## Existing

Original ticket: https://github.com/SEMOSS/semoss-ui/issues/819

### Requirements (What)

- [ ] Add primary-black to the theme.ts file (I think this will have to be added as a custom palette but will leave that to the devs to determine)
Here are the hex codes: Image

bannaarisamy-shanmugham-kanini:
Creating a new custom palette in theme.ts file for specifically buttons

## Description
Adding primary black color option with three type of colors like dark, main and light

## Changes Made
Created a new custom palette with the above mentioned color types

## How to Test

1. Need to be used by developer to use this, while creating or working with buttons
2. the buttons should be in the above mentioned color types


neelneelneel:
This should be updating the primary theme token to black and fixing pieces that can get modified. Can you make these changes and submit a PR to https://github.com/SEMOSS/semoss-ui/tree/1096-build-banner-for-app-builder-landing-page

johbaxter:
Closed per designers request @ktnguyen4 @bfekadu21 @ehynd

PR #1309, opened 2025-06-17
Updated the background throughout App

Gowrishankar-Palanisamy:
This PR contains the following changes

   -->Updated the gray as a Paper 1 
   --> Updated the white as a Paper 2
   -->Updated the button color
   -->Updated the ToggleTabsGroup background color

PR #1343, opened 2025-06-23
Added the SpellCheck functionality

### LINKED ISSUE #1258: Explore feasibility of spell check in text blocks

### Description

If possible, it would be ideal if misspelled words could be underlined in a red squiggly line within text blocks within the UI builder.  

The user could then right click on the misspelled word and the replacement options would pop up.  

Requirements: 
- [x] Determine if this is feasible
- [x] Implement fix for only text blocks  

### Relevant Feature or Code

_No response_

### Requirements (what)

_No response_

Gowrishankar-Palanisamy:

  -->Added one props in QueryInputSettings file.
  -->Updated that prop in text config file

### DIFF
--- a/libs/renderer/src/components/block-settings/custom/QueryInputSettings.tsx
+++ b/libs/renderer/src/components/block-settings/custom/QueryInputSettings.tsx
@@ -65,6 +65,11 @@ interface QueryInputSettingsProps<D extends BlockDef = BlockDef> {
      * Default path map by default {}
      */
     defaultPathMap?: any;
+
+    /**
+     * Spell check for the input
+     */
+    spellCheck?: boolean;
 }
 
 interface Option {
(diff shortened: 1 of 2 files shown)


libs/renderer/src/components/block-settings/custom/QueryInputSettings.tsx
```diff
@@ -609,6 +615,7 @@ export const QueryInputSettings = observer(
                                 overflowX: "auto",
                                 scrollBehavior: "smooth",
                             },
+                            spellCheck: spellCheck ?? false,
```
j-adam-connor, line 618: Should we default the value of `spellCheck` upstream of this?

Gowrishankar-Palanisamy:
This PR changes linked to another PR #1309

https://github.com/SEMOSS/semoss-ui/pull/1309

PR #1301, opened 2025-06-16
Update Files Tab Left Panel UI with Folder/File Actions and Multi-Select Support

Vishal24-Kanini:
## Description
Implemented UI updates for the Files tab left panel, including nested folder support, file/folder actions (rename, duplicate, delete), modal confirmation for deletions, multi-select via checkboxes, and hover tooltips for icons as per the latest design.

## Changes Made
- Updated the left panel menu UI for the Files tab based on the provided design.

- Implemented nested folder structure with support for multiple files under each folder.

- Added 3-dot menu with options to Rename, Duplicate, and Delete files/folders.

- Integrated modal confirmation on delete action as per Figma design.

- Enabled multi-select functionality for files/folders using checkboxes.

- Implemented hover tooltips for all action icons based on Figma specifications.

PR #1345, opened 2025-06-23
Add "Copy to Clipboard" Action for Button OnClick to Copy Variable Value

### LINKED ISSUE #1004: Add ability to expose a copy button to copy from the notebook
### Description 

Add ability to allow app creator to create a button that copies text from notebook / variable to clipboard

### Existing

`ButtonBlock.tsx`

### Requirements

We want to add an event to the onClick for button. 

These are the existing options right now: 

Image

- [x] Add an option for "Copy to Clipboard"
- [x] When this is selected, a second dropdown appears with "Variable".  When the user clicks this dropdown, all their existing variables will show up. 
- [x] When this action is set and the button is clicked, the variable value will be copied to the clipboard.

Vishal24-Kanini:
## Description
This PR introduces a new option to the existing button onClick actions in the app builder interface. App creators can now configure a button to copy the value of a variable to the clipboard when clicked.

## Changes Made
- Added a new onClick action: "Copy to Clipboard".
- When this action is selected:
- A secondary dropdown appears to allow the user to select from existing variables.
- On button click, the selected variable’s value is copied to the clipboard.
- Clipboard functionality is implemented securely and works across supported browsers.

### DIFF
--- a/libs/renderer/src/components/block-settings/ListenerActionOverlay.tsx
+++ b/libs/renderer/src/components/block-settings/ListenerActionOverlay.tsx
@@ -223,6 +223,10 @@ export const ListenerActionOverlay = observer(
                 });
             } else if (message === ActionMessages.DISPATCH_OUTPUTS_EVENT) {
                 setValue("payload", {});
+            } else if (message === ActionMessages.COPY_TO_CLIPBOARD) {
+                setValue("payload", {
+                    text: "",
+                });
             } else if (message === ActionMessages.RUN_CELL) {
                 if (listeners[listener].order[actionIdx]) {
                     if (
@@ -252,6 +256,7 @@ export const ListenerActionOverlay = observer(
             }
         }, [message]);
 
+        console.log(cells)
         return (
             <>
                 <Modal.Title>
@@ -277,6 +282,7 @@ export const ListenerActionOverlay = observer(
                                             ActionMessages.DISPATCH_EVENT,
                                             ActionMessages.DISPATCH_OUTPUTS_EVENT,
                                             ActionMessages.DISPATCH_OPEN_EVENT,
+                                            ActionMessages.COPY_TO_CLIPBOARD,
                                         ].map((a, aIdx) => (
                                             <Select.Item key={aIdx} value={a}>
                                                 {ACTIONS_DISPLAY[a]}
@@ -369,7 +375,7 @@ export const ListenerActionOverlay = observer(
                                                     const variableName =
                                                         state.getAlias(
                                                             queryId,
-                                                            c.id,
+                                                            c.id
                                                         );
 
                                                     return (
@@ -377,13 +383,13 @@ export const ListenerActionOverlay = observer(
                                                             key={c.id}
                                                             value={c.id}
                                                         >
-                                                            <Typography
-                                                                variant={
-                                                                    "body2"
-                                                                }
-                                                            >
-                                                                {variableName}
-                                                            </Typography>
+                                                                <Typography
+                                                                    variant={
+                                                                        "body2"
+                                                                    }
+                                                                >
+                                                                    {variableName}
+                                                                </Typography>
                                                         </Select.Item>
                                                     );
                                                 })}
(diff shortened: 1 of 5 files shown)


libs/renderer/src/components/block-settings/ListenerActionOverlay.tsx
```diff
@@ -252,6 +256,7 @@ export const ListenerActionOverlay = observer(
             }
         }, [message]);
 
+        console.log(cells)
```
j-adam-connor, line 259: Hello! Any reason why we're keeping this log here?

libs/renderer/src/components/block-settings/ListenerActionOverlay.tsx
```diff
@@ -534,6 +540,34 @@ export const ListenerActionOverlay = observer(
...
+                                        >
+                                            {Object.keys(state.variables).map(variableName => (
+                                                <Select.Item key={variableName} value={variableName}>
+                                                    {variableName}
+                                                </Select.Item>
+                                            ))}
+                                        </Select>
+                                    )}
+                                />
+                            </>
```
j-adam-connor, line 569: Minor nitpick but this fragment isn't necessary here.

libs/renderer/src/store/state/state.store.ts
```diff
@@ -547,6 +551,10 @@ export class StateStore {
             } else if (ActionMessages.DISPATCH_OUTPUTS_EVENT === action.message) {
 
                 this.dispatchOutputsEvent()
+            } else if (ActionMessages.COPY_TO_CLIPBOARD === action.message) {
+                const { text } = action.payload;
+
+                this.copyVariableToClipboard(text);
```
j-adam-connor, line 557: We have a TODO comment at the declaration of this function indicating that it should be incorporated with `useBlock`. Is this a good opportunity for that?

libs/renderer/src/components/block-settings/ListenerActionOverlay.tsx
```diff
@@ -369,21 +375,21 @@ export const ListenerActionOverlay = observer(
...
-                                                            >
-                                                                {variableName}
-                                                            </Typography>
+                                                                <Typography
+                                                                    variant={
+                                                                        "body2"
+                                                                    }
+                                                                >
+                                                                    {variableName}
+                                                                </Typography>
```
j-adam-connor, line 392: Looks like these white spaces are formatting errors.

j-adam-connor:
Hello, I just had a few minor suggestions. Also, there are some merge conflicts, so have a look at those please.

libs/renderer/src/components/block-settings/ListenerActionOverlay.tsx
Vishal24-Kanini: Hi @j-adam-connor, I was thinking someone has kept it purposely but as you mention I removed it and committed my changes again.

libs/renderer/src/components/block-settings/ListenerActionOverlay.tsx
Vishal24-Kanini: Removed the fragment as suggested. Thanks for pointing it out!

libs/renderer/src/store/state/state.store.ts
Vishal24-Kanini: The functionality has already been incorporated with useBlock. You can see it in useBlock.tsx—it's handled similarly to existing functionalities like "run query" and "run cell", which are dispatched as listener actions from within useBlock.

libs/renderer/src/components/block-settings/ListenerActionOverlay.tsx
Vishal24-Kanini: yes they are formatting error, I have removes those error also.

Vishal24-Kanini:
Hi, 
All the new suggested changes are moved inside the mentioned PR :- https://github.com/SEMOSS/semoss-ui/pull/1301

PR #1380, opened 2025-06-25
339 ability to add multiple documents when creating a faiss vector

### LINKED ISSUE #339: Ability to add multiple documents when creating a FAISS vector
Right now, at creation step of a FAISS vector, you can only load one document and then you have to go to the the Files tab to load the rest. Within the Files tab, you can load multiple at a time...

This feels like unnecessary steps if the user already has a list of documents to go that need to be vectorized-- why not allow them to upload them all at time of creation? If there is a reasoning behind it, maybe we can at least communicate why we do one document only to start.

SuryaNarayanan-Kgs:
## Description

This PR is Raised for multiple file upload in model section

## Changes Made

Changes are made according to the model page design where , while submitting the user can able to upload multiple file where now it is sent as a array with a comma separator and add model page has some new design with separate section ( select hosting mode , credentials , security ,general 

## How to Test

1. Click on model -> add model -> select a model -> enter the required fields -> in file upload add multiple files -> submit
2. once the model is been submit it will take some time to embed and later navigated to the submitted model . there user can see different tabs in which under files tab you can see uploaded files .

## Notes

NOTE : since except FAISS model , I have to know other models required field so it can be tested on all models and provided seamless delivery


packages/client/src/components/import/ImportForm.tsx
```diff
@@ -470,7 +479,7 @@ export const ImportForm = (props) => {
         if (steps[1].id.includes('File Uploads')) {
             if (steps[1].title === 'ZIP') {
                 const upload = await monolithStore.uploadFile(
-                    [data.ZIP],
+                    data.ZIP,
```
AAfghahi: is this already a zip of files? Based on the uploadFile function, it requires that you pass in an array

packages/client/src/components/import/ImportForm.tsx
```diff
@@ -706,518 +716,1155 @@ export const ImportForm = (props) => {
         }
     };
 
+    console.log('Default Fields:', defaultFields);
```
AAfghahi: are these console logs required?

packages/client/src/components/import/ImportForm.tsx
```diff
@@ -706,518 +716,1155 @@ export const ImportForm = (props) => {
...
-                                                            : ''
-                                                    }
-                                                    onChange={(value) =>
-                                                        field.onChange(value)
-                                                    }
-                                                    inputProps={{
-                                                        'data-testid': `importForm-textField-${val.fieldName}`,
+            {/* Logic flags */}
+            {(() => {
+                const hasOnlyUploadFields = defaultFields.every(
```
AAfghahi: could we have these functions live outside of the form, and then invoke them in the form?

packages/client/src/pages/import/ImportConnectionPage.tsx
```diff
@@ -147,13 +147,17 @@ export const ImportConnectionPage = () => {
 
                 if (values.secondaryFields['EMBEDDINGS']) {
                     const upload = await monolithStore.uploadFile(
-                        [values.secondaryFields['EMBEDDINGS']],
+                        values.secondaryFields['EMBEDDINGS'],
```
AAfghahi: same thing here, unsure why we deleted the array from this pixel call

packages/client/src/pages/import/ImportConnectionPage.tsx
```diff
@@ -147,13 +147,17 @@ export const ImportConnectionPage = () => {
...
 
+                    const fileLocations = upload.map(
+                        (file) => file.fileLocation,
+                    );
+
                     const secondaryPixel = `CreateEmbeddingsFromDocuments(
-                        engine="${output.database_id}", 
-                        filePaths=["${upload[0].fileLocation}"]
+                    engine="${output.database_id}",
+                    filePaths=${JSON.stringify(fileLocations)}
```
AAfghahi: this also might just be an array

AAfghahi (changes requested):

j-adam-connor:
Hello, just adding here, I think we mentioned last week or so that we want to have these description templates filled out. We may adjust the templates in the future, but it's good for us to have a formatted record of changes from the developer. Thank you!

packages/client/src/components/import/ImportForm.tsx
```diff
@@ -15,12 +15,21 @@ import {
...
+    Divider,
+    MenuItem,
+    FormControl,
 } from '@semoss/ui';
-import { ExpandLess, ExpandMore, Help } from '@mui/icons-material';
-
 import { useStepper, useRootStore } from '@/hooks';
 import { useNavigate } from 'react-router-dom';
 import { useFieldArray, useForm, Form, Controller } from 'react-hook-form';
+import React from 'react';
```
j-adam-connor: We're already importing from react at the top of this file. Not sure we need to import `React` either.

packages/client/src/components/import/ImportForm.tsx
```diff
@@ -15,12 +15,21 @@ import {
...
+import { ExpandLess, ExpandMore } from '@mui/icons-material';
 
 const StyledFlexEnd = styled('div')(({ theme }) => ({
```
j-adam-connor: We don't appear to be using this anywhere. Nor are we using `StyledProperty`, `StyledDropzoneField`, or `StyledSubmitButton` below anywhere either.

packages/client/src/components/import/ImportForm.tsx
```diff
@@ -706,518 +716,1155 @@ export const ImportForm = (props) => {
...
+
+                        {/* General Section */}
+                        <div style={{ display: 'flex', flexDirection: 'row' }}>
+                            <div
+                                style={{
+                                    display: 'flex',
+                                    flexDirection: 'column',
+                                    width: '35%',
+                                }}
+                            >
```
j-adam-connor: Curious as to why we're using inline styles here. Some of these components are styled and others are inline. We should try to be consistent in our approach here.

I'd also add that top-to-bottom is the default layout algorithm direction (block). Typically when you find yourself using `flex-direction: column;` you could be using block. I bring this up because flex changes the behavior of its children; for example, it disables margin collapsing.

packages/client/src/components/import/ImportForm.tsx
```diff
@@ -706,518 +716,1155 @@ export const ImportForm = (props) => {
...
+                                >
+                                    Please provide the name, type, and model to
+                                    uniquely identify,
+                                    <br />
+                                    categorize, and configure your setup for
+                                    optimal performance.
+                                </Typography>
+                            </div>
+
+                            <div style={{ width: '65%' }}>
```
j-adam-connor, line 891: Reiterating what was mentioned above here. Is 65% what we want in all cases?

packages/client/src/components/import/ImportForm.tsx
j-adam-connor, line 1: Is there a lot of duplicated code here? Is there an opportunity to extract much of this to a component?

packages/client/src/components/import/ImportForm.tsx
j-adam-connor: To add to this, if we use components for some of the duplicated JSX here, this can be configuration driven instead of hardcoded, which is more brittle.

packages/client/src/pages/import/ImportConnectionPage.tsx
j-adam-connor, line 1: I noticed we have some unused state in this file unrelated to your changes:

```
const [predictDataTypes, setPredictDataTypes] = React.useState(null);
const [metamodel, setMetamodel] = React.useState(null);
```

Removing these will allow you to remove `React` from the imports too, I believe.

packages/client/src/components/import/ImportForm.tsx
SuryaNarayanan-Kgs: While Unzipping the File , the form data is not accepting it as a [object file] getting error as Error while UnZipping, instead it accepting it as (binary). so i have removed and tested it . now its working fine .

packages/client/src/pages/import/ImportConnectionPage.tsx
SuryaNarayanan-Kgs: While Unzipping the File , the form data is not accepting it as a [object file] getting error as Error while UnZipping, instead it accepting it as (binary). so i have removed and tested it . now its working fine .

packages/client/src/pages/import/ImportConnectionPage.tsx
SuryaNarayanan-Kgs: While Unzipping the File , the form data is not accepting it as a [object file] getting error as Error while UnZipping, instead it accepting it as (binary). so i have removed and tested it . now its working fine .

packages/client/src/components/import/ImportForm.tsx
SuryaNarayanan-Kgs: removed

packages/client/src/components/import/ImportForm.tsx
SuryaNarayanan-Kgs: fixed

packages/client/src/components/import/ImportForm.tsx
SuryaNarayanan-Kgs: remoed

packages/client/src/components/import/ImportForm.tsx
SuryaNarayanan-Kgs: removed

packages/client/src/components/import/ImportForm.tsx
SuryaNarayanan-Kgs: fixed

packages/client/src/components/import/ImportForm.tsx
SuryaNarayanan-Kgs: fixed

packages/client/src/components/import/ImportForm.tsx
SuryaNarayanan-Kgs: fixed

SuryaNarayanan-Kgs:
## Description

This PR is Raised for multiple file upload in model section

## Changes Made

Changes are made according to the model page design where , while submitting the user can able to upload multiple file where now it is sent as a array with a comma separator and add model page has some new design with separate section ( select hosting mode , credentials , security ,general 

## How to Test

1. Click on model -> add model -> select a model -> enter the required fields -> in file upload add multiple files -> submit
2. once the model is been submit it will take some time to embed and later navigated to the submitted model . there user can see different tabs in which under files tab you can see uploaded files .

## Notes

NOTE : since except FAISS model , I have to know other models required field so it can be tested on all models and provided seamless delivery

SuryaNarayanan-Kgs:
@j-adam-connor  ,  @AAfghahi  i have Reopened the PR after fixing the comments . please let me know if any changes are need to be resolved .

Test Scrennshot :

image

PR #1383, opened 2025-06-25
feat(code-component): code component updated with figma ui changes

neelneelneel:
## How to Test

2. Expected outcomes


neelneelneel:
- [ ] Code (<code>) is meant to be inline, so it shouldn't take a new line
keep the structure:"inline" flag
- [ ] When Code is wrapped with a CodeContainer, this adds a pre block for the styling
- [ ] Can we tie theme the themeProvider? So it comes via context, not property. Overwrite if it is passed in as a prop

PR #1384, opened 2025-06-25
Database entries  added

ammb-123:
## Description

nonApprovedProductCatalog items to be push in to the DB

## Changes Made

Added ENGINERESTRICTION table in to security DB

## How to Test

Database ,Function,Model,Vector,Storage product items would be restricted as per the business requirement.
[Uploading NonApprovedFromDB.docx…]()


anurag91jain:
Changes taken care in https://github.com/SEMOSS/semoss-ui/pull/1418

PR #1388, opened 2025-06-26
feat(renderer): added frameHeader call for dynamic file render

### LINKED ISSUE #1386: Update Grid to be dynamic

### Description

Current grid assumes we know all the column headers. This is not helpful for returning data where the columns are unknown, such as through the return of a SLP query 

### Relevant Feature or Code

_No response_

### Requirements (what)

- [ ] please see email and ping for all details

Paulson-Robert:
## How to Test

2. Expected outcomes


johbaxter:
Fixed with same code in another PR.

TY Paulson

Issue #1369, opened 2025-06-24
Low Code Select Blocks with Dependencies Do Not Clear On Change


meventura5:
### Is there an existing issue for this?

- [x] I confirm that I have not found an existing or similar issue.

### How to reproduce the issue

1. Create a Drag and Drop app
2. Drag in two select blocks
3. In the notebook, populate the cells so that the second select block is filtered based on the first select block's user input
4. Add On Change processes to the blocks to run the cells again when the block values are changed
5. Select values in both blocks
6. Change the first select block's value
7. See bug in second select block. The previous value is still there although it is no longer an option if you click into the block
### Expected Result

I expected the second select block to clear since it should no longer have the previous value due to changing the first block's value.

### Actual Results

The second select block does not clear although the previous value is no longer in the select block options.

**Example:**
<img width="417" alt="Image" src="https://github.com/user-attachments/assets/7eafc6ac-706f-4795-a0f5-102afe489836" />

The above screenshot shows the initial selections in the select blocks with the Activity dropdown being dependent on the previous Project dropdown.

<img width="257" alt="Image" src="https://github.com/user-attachments/assets/f802caab-415b-42ed-b0b5-1d5e4885afe5" />

Then if the Project dropdown is changed, the following dropdown field does not clear although the options are updated when clicked into it.

The following code is tied to each select dropdown:

**Project Select Block Cell Code**
df_filtered_division = filtered_df[filtered_df['DIVISION'] == "{{division-input}}"]

project_list = df_filtered_division['PROJECT'].unique().tolist()
project_list

**Activity Select Block Cell Code**
df_filtered_project = df_filtered_division[df_filtered_division['PROJECT'] == "{{existing-project-add}}"]
df_filtered_project

activity_list = df_filtered_project['ACTIVITY'].unique().tolist()
activity_list

### Additional Context

We would like this issue resolved as soon as possible since this is a roadblock for deploying this form for our client.

### Requirements or Tasks

- [ ] Clear or reset select blocks if the options change and the previous value is no longer there on change

### Have you reproduced the issue with at least the latest released version?

- [x] Yes

### I have checked my terminal/console and have included relevant text here?

- [x] Yes

alaleung:
so some cases to think about.
1. user wants dependent drop down selected value to clear when data changes.

independent select drop down = isdd
dependent select drop down = dsdd
case 1: no selected value for isdd
1. user selects value for isdd
2. dsdd value is nothing and dsdd option changes

[...]

PR #1415, opened 2025-07-01
fix(ui): Low Code Select Blocks with Dependencies Do Not Clear On Change #1369

alaleung:
## Description
Low Code Select Blocks with Dependencies Do Not Clear On Change #1369

## Changes Made
fix(ui): fix dropdown not clearing when dropdown options are refreshed and value does not exist in option


johbaxter:
Not a fan of solution, lets wait till this issue comes up again which it likely will soon

PR #1418, opened 2025-07-02
ReviewComments fix with latest code

ammb-123:
## Description

nonApprovedProductCatalog items to be push in to the DB

## Changes Made

Added ENGINERESTRICTION table in to security DB

## How to Test

Database ,Function,Model,Vector,Storage product items would be restricted as per the business requirement.
[Uploading NonApprovedFromDB.docx…]()


johbaxter:
We are rewriting the import flow.  Please add an issue on community board.  And reach out to me when requirements are there.

PR #1433, opened 2025-07-07
feat(client/renderer): create notebook variables tie to community block

### LINKED ISSUE #1340: Create Notebook / Community Block Tie

### Description

Right now, our community block flow is reusable UI blocks that get saved into the database.  However, we want reusable components in general, not just reusable UI components.  

The goal is to make a reusable form.  This is just for POC.  There will be a singular notebook dependency tied into the UI.  

### Relevant Feature or Code

Possible Implementation Strategy 

Component Definition:

Image

Dependency Resolution Workflow: 

Image

Possible Steps: 
Saving Components - 
1. Dependency Traversal
- Scan all blocks' data attributes for {{variable}}
patterns
-Parse event listeners for notebook/cell references
Stop at direct dependencies ie (don't traverse notebook's dependencies)
2. Conflict Prevention
- Prefix variables: comp_123_var_a instead of var_a
- Namespace notebooks: comp_123_notebook_1

Loading Components: 
1. Conflict Resolution

Image

2. Template Application 
- Replace all {{var}} references in blocks using mapping table
- Update event listeners with new notebook/cell IDs
- Inject "Community Source" tag into notebook metadata

### Requirements (what)

_No response_

Paulson-Robert:
…cks FE implementation

## Description
Right now, our community block flow is reusable UI blocks that get saved into the database. However, we want reusable components in general, not just reusable UI components.

The goal is to make a reusable form. This is just for POC. There will be a singular notebook dependency tied into the UI.

On this implementation, the dependency variables and notebook/query will brought in along with the community blocks drops.

## Changes Made

1. Before saving the community block, written a scanner function to identify the dependent variable and queries to the block and save that as well.
2. On drop area before generating the desired ui block, trying generate all the dependent quires and variables required and update the block json with the newly generated dependent item's id.
3. Added **isCommunity** flag to the schema to refer a particular block is a community block or not. This flag is also used to isolate the added logic specific to community block.

## How to Test

1. Create a community block with some dependent variables and queries in it.
2. Drag and drop from the community bocks tab, you can notice in notebook and variables panel with some auto generated required variables and notebooks.

[SEMOSS (2).webm](https://github.com/user-attachments/assets/2403d1c3-112a-471d-a791-4d5629ce61d3)


johbaxter:
@KirthikaKumar-K can we merge your community block design changes into this branch

KirthikaKumar-K:
> @KirthikaKumar-K can we merge your community block design changes into this branch

sure @johbaxter

Paulson-Robert:
> @KirthikaKumar-K can we merge your community block design changes into this branch

But @johbaxter her ticket has only FE changes with incomplete features associated with incomplete BE pieces. If we merge it, 1340 ticket can be closed only after the whole implementation. Is it ok?

johbaxter:
Okay yeah lets hold off merging then @KirthikaKumar-K.  We can handle your changes independently.

Do you mind linking your branch in a comment in this thread @KirthikaKumar-K

KirthikaKumar-K:
> Okay yeah lets hold off merging then @KirthikaKumar-K. We can handle your changes independently.
> 
> Do you mind linking your branch in a comment in this thread @KirthikaKumar-K

Sure @johbaxter ,
Here is the [PR](https://github.com/SEMOSS/semoss-ui/pull/1576).
This is the [Branch](https://github.com/SEMOSS/semoss-ui/tree/1402_communityblock_ux_update).

johbaxter:
1. Create app and drop the json saved to this comment.

[CONTAINER_TO_COMMUNITY.json](https://github.com/user-attachments/files/21517757/CONTAINER_TO_COMMUNITY.json)

image

I want to make this container-1 a community block

Looks like we save the data structure to the db correctly, even though i have some thoughts on a refactor of how we save.

But cool now its saved in the community blocks.

3. Create a new app and drag in community block

image

You'll notice some of the run cell actions don't persist when dragging in the community block. As well as no variable/id in the 2nd notebook cell that got dropped in.  Is it due to how we saved it?

Outside of this it does seem the variables do port over.

We will need to take another look at how we do id reassign later on, but this bug fix will need to be done before then (prototype still)


---

Along with this i will need you to test dragging in the community block, when there is already established blocks and notebooks in an app

1.  Make a template app for "Create Diabetes Record"
2. Drag above community block into this (do the ids/variables assign correctly)
3. Make a template app for "Ask LLM"
4. Do step 2 again.

Paulson-Robert:
@johbaxter 
Found the cause and added a fix for the above explained issue - [commit](https://github.com/SEMOSS/semoss-ui/pull/1433/commits/1bab127935b94b6967a084942ae530cf3201805b)

- **Cause** : In the block JSON, as listener which is of **RUN_CELL** type does not have the info about its **alias** **name**. It is referenced as cell id(eg: 1, 2).
- **Fix** : Now I have used the **getAlias** method to find the alias name and include the associated variable to it.

_Yes, the issue is the way we save the json, it has missed some variables._

_I have tested using community along with Template apps with existing vars. Variables and Queries are getting assigned properly.
You can give it a check now._

johbaxter:
We will address a refactor with migration functions at a later point.  Will push this for the purpose of demo.

I have thoughts on the 

1. structure of how we save to db (will blast db in v2)
2. communityBlockMapping saved on blocks
3. id generation prefixed with comm_

libs/renderer/src/store/state/state.store.ts
```diff
@@ -862,6 +876,7 @@ export class StateStore {
 			data: {},
 			listeners: {},
 			slots: {},
+			communityBlockMapping: {},
```
johbaxter, line 879: I don't want this added

libs/renderer/src/store/state/state.actions.ts
```diff
@@ -111,6 +111,7 @@ export interface AddBlockAction extends Action {
                   type: "after";
                   sibling: string;
               };
+        isCommunity?: boolean; // Optional flag to indicate if the block is community-based
```
johbaxter, line 114: Will look at later point but is this needed.

libs/renderer/src/store/state/state.store.ts
```diff
@@ -391,9 +391,9 @@ export class StateStore {
 
 				this.setState(state);
 			} else if (ActionMessages.ADD_BLOCK === action.message) {
-				const { json, position } = action.payload;
+				const { json, position, isCommunity } = action.payload;
 
-				return this.addBlock(json, position);
+				return this.addBlock(json, position, isCommunity);
```
johbaxter, line 396: Wondering if we need isCommunity will explore in a refactor

Paulson-Robert: Need some flag to differentiate in triggering addBlock dispatch for normal block vs community block

PR #1460, opened 2025-07-09
test(renderer): add test for modal block

### LINKED ISSUE #714: Modal Block
### Feature:
Block Defaults

### Component / Function:
The specific component this test will cover.

### Given:

Image

### When:
When an event or user action occurs

### Then:
The expected outcome of this test

stelbailey:
## How to Test

2. Expected outcomes


memisrose (changes requested):
All of the tests pass however I do get the below warning when running:

image

stelbailey:
Just noting here that I opened a [bug ticket](https://github.com/SEMOSS/semoss-ui/issues/1518) for this since it looks like we'd have to modify the actual Modal Block component to get rid of the warning

Issue #721, opened 2025-03-18
Text Block


memisrose:
### Feature:
Block Defaults

### Component / Function:
All Text Blocks including: H1, H2, H3, H4, H5, H6, Text

### Given:

![Image](https://github.com/user-attachments/assets/71016ccc-19f2-4025-bec4-ad6edc057d54)

### When:
When a user drags a text block into the workspace...

### Then:
Then default text should populate along with a font style and size option.

PR #1461, opened 2025-07-09
Text Block #721

alaleung:
## Description
Text block unit test
https://github.com/SEMOSS/semoss-ui/issues/721
## Changes Made
Added Text block unit test

### DIFF
--- a/libs/renderer/src/testing/block-defaults/TextBlock.spec.tsx
+++ b/libs/renderer/src/testing/block-defaults/TextBlock.spec.tsx
@@ -56,8 +56,8 @@ describe("text block", () => {
             blocks: blocks,
         });
 
-        const textBlock = container.querySelector("[data-block='text2']");
-        expect(textBlock.innerHTML).equal("");
+        const textBlock = container.querySelector("[data-block='text2'] p");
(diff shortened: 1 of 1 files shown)


libs/renderer/src/testing/block-defaults/TextBlock.spec.tsx
```diff
@@ -0,0 +1,75 @@
...
+        expect(textBlock).toBeInTheDocument();
+        expect(screen.getByText("Hello world")).toBeInTheDocument();
+    });
+
+    it("does not show", async () => {
+        const { container } = render(<TextBlock id="text2" />, {
+            blocks: blocks,
+        });
+
+        const textBlock = container.querySelector("[data-block='text2']");
```
RNubla: try something along the lines of 
``` ts
const element = screen.queryByRole("progressbar");
expect(element).not.toBeInTheDocument();
```

this will determine if its actually being rendered or not in the DOM

alaleung: Changed

memisrose (approved):
all tests pass with no issues

PR #1466, opened 2025-07-10
Feat/1368 update image block

### LINKED ISSUE #1368: Update Image Block

### Description

Right now, users can only use the image block by putting an external link. These updates mimic the PDF Viewer Block also being updated.

Design ticket: #1319 
Figma: https://www.figma.com/design/Qj4tWPsniFOJ6qSLfxOaBT/Drag-And-Drop---Design?node-id=15729-156054&t=XfZTup7JddzIqTVG-0

### Relevant Feature or Code

_No response_

### Requirements (what)

- [x] Implement Figma Design
There should be 3 option for image selection: 
- [x] Insight: images uploaded by the user of the app when it runs. This dropdown should point to file upload input blocks 
- [x] App: Two options here, the dropdown of all images currently located in the /assets folder and then an option for the user to upload 1 file to the image block.  When uploaded, the file will live in /assets. 
- [x] External: This is the current functionality and points to a URL link

Manikandan-Kanini:
## How to Test

2. Expected outcomes


johbaxter (changes requested):
You add a bunch of changes to config, where is the migration function to add these properites to existing image blocks

johbaxter (approved):

PR #1498, opened 2025-07-11
Add sequencing to new layers (pages & blocks)

### LINKED ISSUE #1396: Add sequencing to new layers

### Description

Blocks should count up numerically based on when they are added to the page.  

Ex: 

Image

I added these in the same order they appear, so I would want them to be named 
/ 
/page--2
/page--3
/page--4

### Relevant Feature or Code

Check out how this is being done in the notebook for reference. 

### Requirements (what)

- [x] Update page ids upon creation to be numeric
- [x] Test old apps on branch to make sure nothing breaks 
- [ ] Use same branch as #1395

Vishal24-Kanini:
## Description
This PR implements the feature, when user add new layers like pages they should count up numerically based on when they are added.
This PR also implements the feature, when user add new blocks they should count up numerically based on when they are added to the page.

## Changes Made
- Added sequencing to the pages, when add a new page, sequencing starts from page--1, page--2, page--3, page--4, and so on.
- Added sequencing to the blocks also when user add new blocks they should count up numerically based on when they are added to the page.
- Sequencing of blocks also starts from block--1, block--2, block--3, even across the multiple pages the count will goes on continuously.
- User can save the app and came again to add new page or a new block, count will where it left.
- Confirmed and put the check that page sequencing should not mix with block sequencing.
- 
image

## How to Test

2. Expected outcomes


libs/renderer/src/store/state/state.store.ts
```diff
@@ -818,7 +818,40 @@ export class StateStore {
      */
     private generateBlock = (json: BlockJSON, parent?: Block["parent"]) => {
         // generate a new id
-        const id = `${json.widget}--${Math.floor(Math.random() * 10000)}`;
+        let id: string;
```
johbaxter: Overall i think we can split this logic out into its own private function.

With this we should probably be a bit more optimal, and not do a loop through blocks and rather do a check.

Will put this in the next commit

PR #1506, opened 2025-07-14
869 scatter plot

### LINKED ISSUE #869: Scatter Plot
### Feature:
The overarching feature this component test is apart of.

### Component / Function:
The specific component this test will cover.

### Given:
Preconditions or current state

### When:
When an event or user action occurs

### Then:
The expected outcome of this test

RNubla:
## Changes Made

* create customRenderHook to be able to reflect changes made via useBlock
* create test cases for scatter plot

## How to Test

1. pnpm run test


johbaxter (approved):
Not sure this actually works, but will push test.  If we have failed test, test team should fix.  But will push for the purpose of adding the test file

PR #1538, opened 2025-07-16
Cron job builder

### LINKED ISSUE #973: Error Creating Custom Cron Jobs

1. Go to the jobs tab and create a new job as shown below
2. Observe that it fails regardless of the entry

Image
Image
Image

### Expected Result

Expected the cron job to schedule just like the regular scheduler does.

### Actual Results

The cron job didn't get set.

### Additional Context

The cron expression looks blatantly wrong in the pixel call, so this might be a UI fix?

### Requirements or Tasks

_No response_

- [x] Yes

- [x] Yes

rithvik-doshi:
Fix the cron job builder modals (see issue for details)


rithvik-doshi:
Note that we're using 7-field cron syntax, which is rather complicated. See url linked in the code for a blog overview of rules and some of the regex. The tricky part is going to be validating between different fields. For example, you can't have specific fields for both day of week and day of month, one of them has to be none if specifying the other.

rithvik-doshi:
There are also some state management issues - the modals aren't getting the jobs properly as I see the default job show up. Nor do I see any of the saved state info

packages/client/src/pages/jobs/JobBuilderModal.tsx
```diff
@@ -257,45 +219,56 @@ export const JobBuilderModal = (props: {
 				}","recipeParameters":""}',triggerOnLoad=[false],triggerNow=[false]);`,
 			);
 			if (response.errors.length) {
-				throw new Error(response.errors[0]);
```
Paulson-Robert, line 258: This logic(already existing one) seems more optimized and it triggers success notification as well. Any specific reason for this change @rithvik-doshi ?

packages/client/src/pages/jobs/JobBuilderModal.tsx
```diff
@@ -257,45 +219,56 @@ export const JobBuilderModal = (props: {
...
+			const response = await runPixel(
+				`META|EditScheduledJob(jobId="${builder.id}",jobName="${
+					builder.name
+				}",${
+					builder.tags.length
+						? `jobTags=${JSON.stringify(builder.tags)},`
+						: ""
+				}jobGroup=["defaultGroup"],cronExpression="${builder.cronExpression}",cronTz="${builder.cronTz}",recipe="<encode>${encode}</encode>",uiState='{"jobType":"${builder.jobType}", "jobName":"${builder.name}", "cronExpression":"${builder.cronExpression}", "cronTimeZone":"${builder.cronTz}"}',triggerOnLoad=[false],triggerNow=[false]);`,
+			);
+			if (response.errors.length) {
```
Paulson-Robert, line 251: Same goes here as well, can throw the error and render the error toast from catch block and return without closing the modal.

packages/client/src/pages/jobs/JobStandardFrequencyBuilder.tsx
```diff
@@ -9,6 +9,9 @@ export const JobStandardFrequencyBuilder = (props: {
 }) => {
 	const { builder, setBuilderField } = props;
 
+	// Flag to prevent circular updates
+	const isParsingCron = useRef(false);
```
antgibson96: I'm not seeing this get plugged into a component, let's shift this over to just a useState so we can drop the current.

antgibson96 (approved):

ehynd:
closing since changes contained in #2221

PR #1541, opened 2025-07-17
build(fix): search bar not clearing on navigation

### LINKED ISSUE #1502: Search Bar not clearing on Navigation

1. On start, search for an already created app/db/function, etc.
2. Navigate away and back, and see that the search is not being cleared

Image

Image

### Expected Result

_No response_

### Actual Results

_No response_

### Additional Context

_No response_

### Requirements or Tasks

_No response_

- [x] Yes

- [x] Yes

JaganSomannaKanini:
## Description
Once we search for anything in the global search, after navigation the search is not cleared. With this PR it is fixed.

## Changes Made
Using the useLocation from the react router, we will be able to get the current route with this we can clear the search. On every route change the search is cleared.

## How to Test

1. On start, search for an already created app/db/function, etc.
2. Navigate away and back.
3. After the search click on the app you want to open, the app opens in a new window. Come back to the window where the search was performed, the search will be cleared.

## Notes

- **Searched for a app** 
image

- **Then navigated to the App/db/function, etc.**

image

- **Searched for an App and clicked on the link to open the App.**

image

**Navigated back to the window where the search was performed, the search is cleared.**

image

### DIFF
--- a/packages/client/src/components/shared/Search.tsx
+++ b/packages/client/src/components/shared/Search.tsx
@@ -1,6 +1,6 @@
-import { useMemo, useState } from 'react';
+import { useEffect, useMemo, useState } from 'react';
 import { observer } from 'mobx-react-lite';
-import { useNavigate } from 'react-router-dom';
+import { useLocation, useNavigate } from 'react-router-dom';
 
 import Autocomplete from '@mui/material/Autocomplete';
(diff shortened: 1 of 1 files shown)


johbaxter (approved):
Looks good to me, ty

PR #1549, opened 2025-07-18
feat(client): check if user want to export data

AJMADHAB:
## Description
Check if user wants to export data alongwith database export

## Changes Made
1. Modified EngineHeader.tsx

## How to Test
1. Steps to reproduce/test the behavior:
    Go to Database Page -> Open any Database -> click on Export button
2. Expected outcomes:
    You will be able to see the popup(displaying that if user wants to export data alongwith database)


packages/client/src/components/engine/EngineHeader.tsx
```diff
@@ -130,14 +134,60 @@ export const EngineHeader: React.FC = () => {
                                         )
                                     }
                                     variant="outlined"
-                                    onClick={() => exportDB()}
+                                    onClick={() => setOpenExportModal(true)}
```
johbaxter: This modal likely does not apply for every other engine outside of database.  Function and Model probably has no data associated to it

johbaxter (changes requested):

packages/client/src/components/engine/EngineHeader.tsx
anurag91jain: @johbaxter that's correct. Additionally it is required only for H2 and MySQL. Will make changes accordingly

AJMADHAB:
Opened new PR : #1766

PR #1555, opened 2025-07-18
Adding Add member button for dependency list in app

### LINKED ISSUE #1546: Ability to give permissions to different engines used within an app

### Description

When I open an app, I should be able to give access to all dependencies in 1 go.

When I go to project dependencies tab in settings of an app, I should be able to select a member, and provide them the access to all the engines that are being used in the app.

### Relevant Feature or Code

_No response_

### Requirements (what)

_No response_

pallabi2303:
## Description
Add Member Button for changing and giving access for dependent engines to members

## Changes Made
1. It will show a Add Member button in the dependency tab 
2. On Click it will open MemberDependencyOverlay to search users
3. on choosing a member, it will show all the engines with labels which have access and which doesn't 
4. User can give access or change access for the engines

## How to Test
image
image
image


pallabi2303:
@CodiumAI-Agent /review

pallabi2303:
@CodiumAI-Agent /improve

PR #1585, opened 2025-07-24
feat(client): jira connector changes

### LINKED ISSUE #1363: Jira Connector UI Changes

### Description

UI changes for developing a new Jira Profile card in the settings page and saving the API keys.

### Relevant Feature or Code

_No response_

### Requirements (what)

_No response_

SindhuHarika:
## Description
Added new Jira Connector

## Changes Made
Added the changes for the jira connector both in the settings page and the drag and drop.


johbaxter (changes requested):
Please utilize the notebook for the business logic of the reactor

The idea is to keep blocks as simple as possible this app will likely have some need of an iterator block and the need to show cards for each Jira ticket.

As for the JiraInsertApi key this would be a button block tied to a cell that runs Jira insertAPI Key.

The blocks are meant to be as simple as possible.

@anurag91jain @SindhuHarika Does this make sense?
JiraGetTickets 

should be in the notebook and you pass a list to this block

johbaxter:
Look at this comment:

https://github.com/SEMOSS/semoss-ui/pull/1597/#issuecomment-3276355571

PR #1586, opened 2025-07-24
Landing Page Changes - Img Changes reverted

ammb-123:
[Uploading LandingPage_Change.docx…]()
## Description

Redesign DT AI Core Landing page.

## Changes Made

Redesign DT AI Core Landing page.
## How to Test

1,Go to Login Page
2.Login will updated with new Img

## Notes

Attached Unit Test Document


neelneelneel (changes requested):
Please load custom images from the theme database directly. Do not push into hte base platform

johbaxter:
This is stale, i think we should have the image stored on backend.

And on the the theme.loginBannerImage we just store the url and point to it on the Frontend

PR #1588, opened 2025-07-24
Update PDF Viewer Block #1260

### LINKED ISSUE #1260: Update PDF Viewer Block

### Description

Right now, the PDF Viewer Block does not allow the user to select a file to actually view, it says "Please contact admin" instead.  

The block is being hidden by #1252 until dev work is complete.

Requirements: 

- [x] Within the block settings menu, the user needs to be able to select a file 
- [ ] Need to be able to pull from insight (ex: user uploads a file), engine (ex: file stored within) or app (ex: saved in the app)

Gowrishankar-Palanisamy:

  -->Updated the PDF viewer block settings with engine , app tab.
  -->Insight tab is not implemented for PDF viewer block settings.
  -->As of now I'm showing files from all the engine catalogs for engine tab.


libs/renderer/src/components/block-defaults/pdfViewer-block/PDFViewerBlock.tsx
```diff
@@ -89,56 +90,54 @@ export const PDFViewerBlock: BlockComponent = observer(({ id }) => {
...
+            if (engineIds) {
+                response = await runPixel<[string]>(
+                    `GetEngineAssetsBase64(filePath=["${path}"], engine=["${engineIds}"]);`,
+                );
+                // Directly use base64 from response
+                const base64Content = response?.pixelReturn[0]?.output;
+                if (!base64Content) throw new Error("Failed to get base64 PDF");
+                // Ensure prefix
+                const pdfPrefix = "data:application/pdf;base64,";
+                return base64Content.startsWith(pdfPrefix)
```
j-adam-connor: We're defining the prefix variable twice. Let's just do it once--and possibly in an external file if there's a possibility that we'd use it elsewhere.

`export const PDF_FILE_PREFIX = "data:application/pdf;base64,";`

Also, we might benefit from a util function. Something like:

```
const normalizePdfBase64 = (content: string): string => {
    if (content.startsWith(PDF_FILE_PREFIX)) {
        return content;
    }
    
    // Extract base64 content, removing any existing data URL prefix
    const base64Content = content.replace(/^data:.*?;base64,/, "");
    return PDF_FILE_PREFIX + base64Content;
};
```

libs/renderer/src/components/block-defaults/pdfViewer-block/PDFViewerBlock.tsx
```diff
@@ -89,56 +90,54 @@ export const PDFViewerBlock: BlockComponent = observer(({ id }) => {
...
-                        );
-                        resolve(pdfPrefix + base64Content);
-                    } else {
-                        resolve(base64data);
-                    }
-                };
-                reader.onerror = reject;
-                reader.readAsDataURL(blob);
-            });
+                        if (!base64data.startsWith(pdfPrefix)) {
```
j-adam-connor: Just echoing the previous comment here.

packages/client/src/components/blocks-workspace/blocks/settings/shared/PDFViewerSettings.tsx
```diff
@@ -29,86 +61,534 @@ interface PDFViewerSettings<D extends BlockDef = BlockDef> {
...
+        const getAssetsApp = useMemo(() =>
+            runPixel<any>(
+                `BrowseAppAssets(project=["${appId}"], filePath=["/"]);`
+            ),
+            [appId]
+        );
+
+        const getEngineId = useMemo(() =>
+            `MyEngines(engineTypes=[${allEngineTypes.map(type => `"${type}"`).join(",")}]);`,
+            [allEngineTypes]
```
j-adam-connor: `allEngineTypes` is a static array and will never change, so there's no point to having it in this dependency array. In fact, since it's a constant defined in the component, it will evaluate on every render, causing this `useMemo` to also fire on every render.

packages/client/src/components/blocks-workspace/blocks/settings/shared/PDFViewerSettings.tsx
```diff
@@ -29,86 +61,534 @@ interface PDFViewerSettings<D extends BlockDef = BlockDef> {
...
+        );
+
+        const engineIDResponse = useMemo(() => runPixel(getEngineId), [getEngineId]);
```
j-adam-connor: First, a minor nitpick. We have divergent naming conventions. `engineId` or `engineID`--let's settle on one.

Second, since  `getEngineId` will never change between renders, there isn't much point to having it in this dependency array. That realization leads us to question the usefulness of storing this Promise in  a memoized constant at all. What does this achieve?

packages/client/src/components/blocks-workspace/blocks/settings/shared/PDFViewerSettings.tsx
```diff
@@ -29,86 +61,534 @@ interface PDFViewerSettings<D extends BlockDef = BlockDef> {
...
+
+        useEffect(() => {
+            // Fetch App Assets
+            getAssetsApp.then((result) => {
```
j-adam-connor: Any reason we use `then` in some places and `try-catch` in others?

packages/client/src/components/blocks-workspace/blocks/settings/shared/PDFViewerSettings.tsx
```diff
@@ -29,86 +61,534 @@ interface PDFViewerSettings<D extends BlockDef = BlockDef> {
...
+                        ).values()
+                    )
+                    : [];
+                fetchEngineOptions(engineIds);
+            }).catch((error) => {
+                console.error("Error fetching engines:", error);
+            });
+        }, []);
+
+        const fetchEngineOptions = async (engineIdsList) => {
```
j-adam-connor: Does `engineIdsList` have a type?

packages/client/src/components/blocks-workspace/blocks/settings/shared/PDFViewerSettings.tsx
```diff
@@ -29,86 +61,534 @@ interface PDFViewerSettings<D extends BlockDef = BlockDef> {
...
+                filesByType[type].length > 0
+                    ? filesByType[type]
+                    : [{
+                        id: `${type}-empty`,
+                        path: "",
+                        display: "",
+                        group,
+                    }]
+            );
+        }, [allEngineTypes, engineOptions, groupAliasMapper]);
```
j-adam-connor: Here again we have a static array, `allEngineTypes`, in the dependency array. `groupAliasMapper` is also defined inside of the component, so it calculates on every render. Consider that when you define functions or variables within the component they will evaluate on every render. So if we add these to a `useMemo` dependency array, we are causing the logic within it to also calculate on every render, which defeats the purpose of `useMemo`. We should remove both from this `useMemo` dependency array.

packages/client/src/components/blocks-workspace/blocks/settings/shared/PDFViewerSettings.tsx
```diff
@@ -29,86 +61,534 @@ interface PDFViewerSettings<D extends BlockDef = BlockDef> {
...
+                        setData("engineId", "", true);
+                        setSelectedPdfPath(value?.path || "");
+                    }}
+                    freeSolo={false}
+                    renderInput={(params) => (
+                        <TextField {...params} placeholder="Select File" size="small" variant="outlined" />
+                    )}
+                />
+                <Box sx={{ display: 'flex', justifyContent: 'center', alignItems: 'center', width: '100%' }}>
+                    <Typography sx={{ color: '#808080' }} variant="body2">Or</Typography>
```
j-adam-connor: Do we not have a constant anywhere for this hex?

packages/client/src/components/blocks-workspace/blocks/settings/shared/PDFViewerSettings.tsx
```diff
@@ -29,86 +61,534 @@ interface PDFViewerSettings<D extends BlockDef = BlockDef> {
...
+                                    const allEmpty = optionsForApp.every(option => !option.display || !option.display.trim());
+                                    return (
+                                        <StyledMenuSection key={appName}>
+                                            <StyledMenuSectionTitle expandIcon={<ExpandMore />}>
+                                                <Typography variant="body2" sx={{ pl: 2 }} fontWeight="bold">{appName}</Typography>
+                                            </StyledMenuSectionTitle>
+                                            <Accordion.Content sx={{ borderTop: "1px solid #e0e0e0" }}>
+                                                <List disablePadding>
+                                                    {allEmpty ? (
+                                                        <Typography variant="body2" sx={{ marginTop: "5px", color: "#DA291C", paddingLeft: 4 }}>
```
j-adam-connor: We should be passing `'error.main'` here instead of hardcoding the value.

packages/client/src/components/blocks-workspace/blocks/settings/shared/PDFViewerSettings.tsx
```diff
@@ -29,86 +61,534 @@ interface PDFViewerSettings<D extends BlockDef = BlockDef> {
...
+}) => (
+    <Stack sx={{ paddingLeft: "9px", marginTop: '12px' }}>
+        {uploadFiles ? (
+            <Stack>
+                <Box sx={{ display: 'flex', justifyContent: 'space-between', alignItems: 'center' }}>
+                    <Typography variant="body1">
+                        {uploadFiles.name}
+                    </Typography>
+                    <DeleteOutline
+                        style={{ cursor: 'pointer', color: '#DA291C' }}
```
j-adam-connor: Let's try to be consistent with our styling conventions.

```
<DeleteOutline
    sx={{ cursor: 'pointer', color: 'error.main' }}
    onClick={() => {
        setUploadFiles(null);
    }}
/>
```

packages/client/src/components/blocks-workspace/blocks/settings/shared/PDFViewerSettings.tsx
```diff
@@ -29,86 +61,534 @@ interface PDFViewerSettings<D extends BlockDef = BlockDef> {
...
+                    multiple={false}
+                    value={uploadFiles}
+                    disabled={isLoading}
+                    onChange={(newValue: File) => {
+                        setUploadFiles(newValue);
+                        addFile(newValue);
+                    }}
+                    style={{
+                        marginTop: '10px',
+                    }}
```
j-adam-connor: Let's get rid of this inline style and add a bottom margin to the sx object of `Box` above. Consider this a prompt to generally avoid inline styles in this PR as well, please.

j-adam-connor:
Hello! Nice work! I did confirm that the the basic functionality is working as expected. I've left some comments just around code quality. In addition to my comments, please take a look at the review agent's suggestions. Let me know if you have any questions or need support!

Gowrishankar-Palanisamy:
Hi @j-adam-connor ,

Addressed your review comments , Thanks

packages/client/src/components/blocks-workspace/blocks/settings/shared/PDFViewerSettings.tsx
```diff
@@ -14,6 +14,8 @@ import { Stack, Box, TextField, Tabs, styled, FileDropzone, Accordion, Typograph
 
 import { ExpandMore, InfoOutlined as InfoIcon, DeleteOutline } from "@mui/icons-material";
 
+const GRAY_COLOR = "#808080"; // Declare common gray color
```
j-adam-connor: Since we use this color in other areas of the app, we should probably not declare it locally. Should we add this to `theme.ts`?

packages/client/src/components/blocks-workspace/blocks/settings/shared/PDFViewerSettings.tsx
```diff
@@ -43,15 +45,41 @@ const StyledMenuSectionTitle = styled(Accordion.Trigger)(({ theme }) => ({
...
+    paddingLeft: 32
+}));
+
+const StyledAutocomplete = styled(Autocomplete)(({ theme }) => ({
+    paddingLeft: "9px",
+     marginTop: '12px' 
+}));
+
+const StyledEngineContent = styled(Accordion.Content)(({ theme }) => ({
+    borderTop: "1px solid #e0e0e0",
```
j-adam-connor: Not sure if I just didn't catch this before but `#e0e0e0` is `background` in `theme.ts`.

packages/client/src/components/blocks-workspace/blocks/settings/shared/PDFViewerSettings.tsx
```diff
@@ -43,15 +45,41 @@ const StyledMenuSectionTitle = styled(Accordion.Trigger)(({ theme }) => ({
...
+}));
+
+const StyledTypographyEror = styled(Typography)(({ theme }) => ({ 
+    color: theme.palette.error.main,
+    paddingLeft: 32
+}));
+
+const StyledAutocomplete = styled(Autocomplete)(({ theme }) => ({
+    paddingLeft: "9px",
+     marginTop: '12px'
```
j-adam-connor: There appears to be extraneous white space here. Also, we have double quotes mixed with single quotes, and we're missing the trailing comma. Please make sure your auto-formatting is respecting prettierrc.

packages/client/src/components/blocks-workspace/blocks/settings/shared/PDFViewerSettings.tsx
```diff
@@ -43,15 +45,41 @@ const StyledMenuSectionTitle = styled(Accordion.Trigger)(({ theme }) => ({
...
 }));
 
+const StyledDeleteIcon = styled(DeleteOutline)(({ theme }) => ({
+    color: theme.palette.error.main,
+    cursor: "pointer",
+}));
+
+const StyledTypographyEror = styled(Typography)(({ theme }) => ({ 
+    color: theme.palette.error.main,
+    paddingLeft: 32
```
j-adam-connor: This value should be a string and have a trailing comma, I think. Make sure you have auto-formatting set up and that it's respecting prettierrc rules.

libs/renderer/src/components/block-defaults/pdfViewer-block/PDFViewerBlock.tsx
```diff
@@ -19,7 +19,7 @@ export interface PDFViewerBlockDef extends BlockDef<"pdfViewer"> {
             padding: string;
         };
         selectedPdf: string | null;
-        engineId?: string;
+        engineID?: string;
```
j-adam-connor: I apologize for the confusion, but the rest of the project uses `engineId` and you use `insightId` here too. So let's just use `engineId`.

packages/client/src/components/blocks-workspace/blocks/settings/shared/PDFViewerSettings.tsx
```diff
@@ -125,50 +152,56 @@ export const PDFViewerSettings = observer(
             [appId]
         );
 
-        const getEngineId = useMemo(() =>
+        const getEngineID = useMemo(() =>
             `MyEngines(engineTypes=[${allEngineTypes.map(type => `"${type}"`).join(",")}]);`,
-            [allEngineTypes]
+            []
         );
```
j-adam-connor: Can you explain to me why you're memoizing this?

packages/client/src/components/blocks-workspace/blocks/settings/shared/PDFViewerSettings.tsx
```diff
@@ -125,50 +152,56 @@ export const PDFViewerSettings = observer(
...
 
-        const engineIDResponse = useMemo(() => runPixel(getEngineId), [getEngineId]);
+        const engineIDResponse = useMemo(() => runPixel(getEngineID), []);
```
j-adam-connor: Same question as above here. Can you please explain why you chose to memoize this?

packages/client/src/components/blocks-workspace/blocks/settings/shared/PDFViewerSettings.tsx
```diff
@@ -125,50 +152,56 @@ export const PDFViewerSettings = observer(
...
+                            }))
+                        : [];
+                    setAppOptions(outputNames);
+                } catch (error) {
+                    console.error("Error fetching assets:", error);
+                }
+        
+                // Fetch Engine IDs
+                try {
+                    const result = await engineIDResponse;
```
j-adam-connor: I can't see any reason for this. Can you help me understand why we would want to store `engineIDResponse` in a memoized variable instead of just invoking `runPixel()` here?

packages/client/src/components/blocks-workspace/blocks/settings/shared/PDFViewerSettings.tsx
```diff
@@ -125,50 +152,56 @@ export const PDFViewerSettings = observer(
...
+        
+            fetchAssetsAndEngines();
         }, []);
 
-        const fetchEngineOptions = async (engineIdsList) => {
-            let pdfFiles: { file_name: string; app_type: string; file_type: string; file_path: string; engineId: string; app_name: string; }[] = [];
-            const getFiles = engineIdsList.map((id) => ({
+        const fetchEngineOptions = async (engineIDsList: EngineIDType[]) => {
+            let pdfFiles: { file_name: string; app_type: string; file_type: string; file_path: string; engineID: string; app_name: string; }[] = [];
+            const getFiles = engineIDsList.map((id) => ({
```
j-adam-connor: Instead of changing everything to `ID`, we should just have changed the few inconsistent instances to `Id`.

j-adam-connor:
Thank you for the quick turnaround on this! I've left some follow-up comments for you. Please make sure you address the issues raised by the code review agent as well.

Gowrishankar-Palanisamy:
Hi @j-adam-connor ,
Addressed your review comments

packages/client/src/components/blocks-workspace/blocks/settings/shared/PDFViewerSettings.tsx
Gowrishankar-Palanisamy: I checked it, there is no need of memoizing , so I removed and moved that into initial useEffect,
Thanks

Abhisheksaini129QA:
@Gowrishankar-Palanisamy , I have tested the above task and have couple of observation related to task, could you please review it.
link - https://github.com/SEMOSS/semoss-ui/issues/1684
CC- @rameshpaulraj

PR #1597, opened 2025-07-25
feat(client): Gmail block connector FE

pallabi2303:
## Description
Gmail Block Component For Drag & Drop Apps

## Changes Made
Created Block Component for implementing features send mail, read mail, delete mail, get sent email List, get all emails, get unread emails.

## How to Test
image
image
image
image
image

PR #1600, opened 2025-07-28
feat(client): spreadsheet connector changes

### LINKED ISSUE #1416: Create connector for Google Spreadsheets - FE

### Description

Create connector for Google Spreadsheets for below functionality :
1.Create and save spreadsheets profile
2.Create new spreadsheets
3.Edit existing spreadsheets
4.Read existing spreadsheets
5.Delete existing spreadsheets

### Relevant Feature or Code

_No response_

### Requirements (what)

_No response_

SindhuHarika:
## Description
Spreadsheet Connector Changes.

## Changes Made
Added the changes for the spreadsheet connector in drag and drop section.


anurag91jain (approved):

libs/renderer/src/components/block-defaults/spreadsheet-block/SpreadsheetBlock.tsx
```diff
@@ -0,0 +1,563 @@
+import React, { CSSProperties, useEffect, useState } from "react";
+import { observer } from "mobx-react-lite";
```
johbaxter, line 2: Can we organize imports

- 3rd party libs

- internal libs `@semoss/ui` @semoss/sdk`

- modules internal to the package/lib we are coding in

libs/renderer/src/components/block-defaults/spreadsheet-block/SpreadsheetBlock.tsx
```diff
@@ -0,0 +1,563 @@
...
+    slots: never;
+    listeners: never;
+}
+
+export const SpreadsheetBlock: BlockComponent = observer(({ id }) => {
+    const { data, setData} = useBlock<SpreadsheetBlockDef>(id);
+    const [isDelete, setIsDelete] = useState(false);
+    const [allListedSheets, setAllListedSheets] = useState([]);
+    const notification = useNotification();
+    const [loggedInUser,setLoggedInUser]= useState('');
```
johbaxter, line 56: code format

libs/renderer/src/components/block-defaults/spreadsheet-block/SpreadsheetBlock.tsx
```diff
@@ -0,0 +1,563 @@
...
+            }
+        } catch (error: any) {
+            notification.add({
+                color: "error",
+                message: error.message,
+            });
+        }
+    };
+
+    const handleUpdateSheet= async (titleSheetId,sheetId) =>{
```
johbaxter: format

johbaxter (commented):
Alright so this block is very tailored to a specific use, meaning we have a lot of logic baked into this block that is specific to Google.

`GoogleGetAllSheets`

The idea for these blocks is to be as generic as possible and that is the whole reason for the notebook.  Your business logic should live in the notebook, and the reactors you have in the block line up with that criteria, of business logic.

GoogleGetAllSheets
GoogleUpdateSheet
GoogleReadSheet
and etc...

So how do you interact with these reactors via the notebook.  You should be able to do so via the event listeners on the block.

The idea of drag and drop is for things to be configurable. 

Not hardcoded.

When you hardcode a button in the block, the user cannot configure much on that button let alone move it around since it is baked into the block code.

Think about this when you create new blocks.  I know it solves your tailored use case but, think about it more in the terms of generic behavior and user customization

johbaxter:
https://github.com/SEMOSS/semoss-ui/pull/1597/#issuecomment-3276355571

Look at comment ^

PR #1647, opened 2025-08-05
Removed duplicate components whose names ending with two

### LINKED ISSUE #1571: Remove extra components from @semoss/ui

### Description

Remove extra components from @SEMOSS/usi 

### Relevant Feature or Code

`/libs/ui/src/index.ts`

### Requirements (what)

- [x] Remove Any component that ends in Two. Components that are exported should be used.

Gowrishankar-Palanisamy:

   --> Removed the duplicated components whose name were ending with "Two".
   --> Updated the common components as per the requirement.


johbaxter (approved):

johbaxter:
image

You removed quite a bit of them as was intended.  But we left a couple out that we should probably make the component files for.

Create semoss/ui components for leftovers in index.ts

PR #1687, opened 2025-08-12
Refactor and checked rest Calls for CSRF support

### LINKED ISSUE #1536: Check Rest Calls for CSRF support

### Description

We need to look at the newest changes that were pushed and make sure the Rest Calls are done in a way to ensure it supports CSRF if it is enabled.  

### Relevant Feature or Code

Issue: 

Image

Should be able to enable it by uncommenting out webxml filter in Monolith project 

Image

### Requirements (what)

_No response_

Vishal24-Kanini:
## Description
This PR addresses CSRF protection support for REST API calls. Every runPixel call is supposed to get attach with CSRF.

## Problem
- I have found that the current interceptor setup in base.ts, both request GET and POST types using the same interceptor logic, even during the CSRF token retrieval process. i am not sure whether this causing the infinite loop or something else because request is hitting the backend but token didn't attach to the further runPixel call.
- i also found that the in App.tsx and config.store, setGeneneralReactore() will run before ENV.CSRF became true when component mounts in App.tsx and that's why CSRF didn't get attached to it and META|Helpjson() getting failed.

image

## Changes Made
- bypass the interceptor when fetching the CSRF token so it doesn’t trigger itself. Used raw fetch(), not the wrapped get().
- I have set CSRF flags inside initialize() before setGeneralReactors() in config.store.
- in this way the csrf is working , CSRF should support every rest call when enabled.

image


johbaxter:
We will need to refactor some of the logic here.

We make one too many fetchCSRF calls.

We likely need to relocate where we do our interceptor, or it may be that the initial Env.update is not synchronously happening with the runPixels that happen on load.

PR #1700, opened 2025-08-13
feat(client): tab block container

### LINKED ISSUE #59: [FEAT] Create a Tab Block

In legacy SEMOSS, a user can place panels behind one another with the tabs. This is important for dashboarding. 

Explore feasibility of a "tab block".  This block would act like a container, but instead of putting blocks next to or below one another, they would be behind each other in tabs.  

Users would need to be able to change the name of the tabs 

### Additional context
Legacy SEMOSS: 
image
Panels are next to each other 

image
Panels can be dragged behind one another with the user able to select which tab is visible on the screen at the top (where the yellow highlight is) 

### Requirements 

- [ ] Explore feasibility of tab block.  If possible, move forward with development.  If not possible, detail why in comments.

SuryaNarayanan-Kgs:
## Description

Tab Block container allows the user to add / edit / delete tab block . It can add multiple tabs and with the Max of 20 Tabs and can be able to change the label name of the Tab.

## Changes Made

Created a new section for Tab Block and create a separate Slot for add multiple slots into a tab .

## How to Test

1. Drag and Drop a Tab Block left section 
2. By default three tabs will be available and can be adding more tabs with a max value of 20 Tabs in the block settings .
3. Label's can be changes for each tab in the block settings .
4. onChange function is also invoked in block settings .

## Notes

If the tab length is excessive ,please use the scroll for viewing the hidden .
The label name can be changed from the selection from the dropdown .
Tab length will be changed on onBlur function ( user needs to enter the no of tabs value and click outside the block) .


johbaxter:
Im curious on the need for the SET_BLOCK_SLOTS. Will review further later.

Issue #1744, opened 2025-08-20
Enable Ability to Login to Multiple Providers


ehynd:
### Is there an existing issue for this?

- [x] I confirm that I have not found an existing or similar issue.

### Description

On the legacy BI tool, users are able to login to multiple providers if enabled. I could login via the platform, but then also login to google, github etc.  This is necessary to use features associated with those accounts. 

On the new platform, this is not possible. We need to enable this multi account login once already in the platform.  

Expected flow:
I login to the platform via native or Microsoft 
I click my name in the bottom left (where the logout is) 
I am able to login to other accounts 

Design ticket: https://github.com/SEMOSS/semoss-ui/issues/1812
Figma: https://www.figma.com/design/qwK4QWfPMnW0lzC6liAy84/Jobs-Scheduler?node-id=496-14433&t=BkiRwMjET3SS5vJO-1

### Relevant Feature or Code

<img width="242" height="200" alt="Image" src="https://github.com/user-attachments/assets/6055a798-006d-4d2b-a5d7-08ce8d737bbd" />

this is where the options lived in the legacy tool 

### Requirements (what)

- [ ] Using social props, show enabled accounts within the login popup
- [ ] Implement Figma for UI design

PR #1758, opened 2025-08-22
Login button block

### LINKED ISSUE #1944: Update the button block to accept login types

### Description

The same button component should be enabled for different login types.

### Relevant Feature or Code

_No response_

### Requirements (what)

_No response_

SindhuHarika:
## Description
Made changes in the button block to include the Login to google changes.

### DIFF
--- a/libs/renderer/src/components/block-defaults/button-block/ButtonBlock.tsx
+++ b/libs/renderer/src/components/block-defaults/button-block/ButtonBlock.tsx
@@ -1,92 +1,124 @@
-import { CSSProperties, useEffect } from "react";
+import { Button, CircularProgress, styled } from "@mui/material";
 import { observer } from "mobx-react-lite";
-
+import { CSSProperties, useEffect } from "react";
+import { oauth } from "@semoss/sdk/react";
+import { useNotification } from "@semoss/ui";
 import { useBlock } from "../../../hooks";
-import { BlockDef, BlockComponent, ListenerActions } from "../../../store";
-
-import { CircularProgress, Button, styled } from "@mui/material";
+import { BlockComponent, BlockDef, ListenerActions } from "../../../store";
 
 const StyledButton = styled(Button, {
-    shouldForwardProp: (prop) => prop !== "loading",
+	shouldForwardProp: (prop) => prop !== "loading",
 })<{ loading?: boolean }>(({ loading }) => ({
-    "& .MuiButton-endIcon svg": {
-        visibility: loading === true ? "hidden" : "visible",
-    },
-    "& .MuiButton-startIcon svg": {
-        visibility: loading === true ? "hidden" : "visible",
-    },
+	"& .MuiButton-endIcon svg": {
+		visibility: loading === true ? "hidden" : "visible",
+	},
+	"& .MuiButton-startIcon svg": {
+		visibility: loading === true ? "hidden" : "visible",
+	},
 }));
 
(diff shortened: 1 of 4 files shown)


libs/renderer/src/components/block-defaults/button-block/ButtonBlock.tsx
```diff
@@ -1,92 +1,124 @@
...
+		}
+	}, []);
+
+	const Login = async (provider) => {
+		try {
+			const response = await oauth(provider);
+			if (response.name) {
+				notification.add({
+					color: "success",
+					message: "Successfully logged into Google",
```
johbaxter, line 75: You have google hardcoded here.

We have other oauth options??

johbaxter (changes requested):
Either way, you should def be able to catch my comment i had before this and not hardcode the provider in the notifications

Who assisted you with the direction on how to go about this, if noone take what i say below into account.

---

But lets think about this differently.

Blocks are meant specifically for UI interaction with data/api.  Not necessarily the api itself.

Notebooks are our API layer.

With that said notebooks should handle our oauth logic, or better yet native login logic as well.

The way i would go about this is, I leave the ButtonBlock as is.

- [ ] create a new notebook cell that does this specific oauth sdk call (with a provider as input on the cell)

And the way the app builder would use this is, when they decide to build their custom login page.  They add a button block and add an onClick event to trigger your new notebook cell, and onSuccess (will have to build this out) we will navigate to another page.

With this said you will probably want a new native login cell that has 2 inputs as well.

- [ ] create native login cell, that has 2 inputs user and pass.

There is a blocker the way our cell execution is done is it takes simply pixel.  We may want to extend our cell configs to execute custom api layers outside of pixel too.

johbaxter:
@anurag91jain and @SindhuHarika 

Also fix conflicts before putting a PR in ready for review....

Move it over to draft if it is not indeed ready (conflicts)

johbaxter:
Also what is ticket # is this tied to?

If it is this one https://github.com/SEMOSS/semoss-ui/issues/1744

You missed the mark as far as the ask.  Even though the feat your proposing in this is valid.

Feel free to send a message to me on Teams if you want more clarity on above

libs/renderer/src/components/block-defaults/button-block/ButtonBlock.tsx
anurag91jain: @johbaxter this doesn't need to be hardcoded. Will change this

anurag91jain:
@johbaxter moved the ticket to draft.

The ticket is not tied to #1744. Reasons why we went with this approach:
1. The initial thought was to create a login button block extending the current button block. This way we would have all the features of button and add a new functionality to it. But not really sure if that's possible with the way our blocks are setup.
2. Hence the idea of adding a type to button came up. The extension to this would be able to select a provider. We can add that feature as well and then keep on adding providers as we implement blocks related to them.
3. The reason why we didn't decouple the notebook call in this case because the oauth process is complex and it would be too much implementation for the app developer (We need to make an API call to check if the user is logged in, otherwise open a window related to the oauth provider, and keep on checking if the window is closed, then check if the user was succesfully logged in). Additionally, it is not a pixel call but an API implementation and currently we don't have the feature to implement API calls. Understandable that we want to decouple the notebooks and frontend, but I feel that would lead to lot of manual effort while trying to create a simple app.

PR #1774, opened 2025-08-26
fix(client): prevent failure when group name starts with symbols

### LINKED ISSUE #1763: Routing is failing if group name starts with /

### Description

2 issues
1. According to keycloak and probably other ADs, the group names start with /. The redirection on FE fails because the URL beccomes //<groupName>
2. When the user lands on the team details page, the name and type were being fetched from state instead of BE. Hence if I bookmark or copy/paste the url in a new tab, I was unable to reach the team details page.
3. When calling /addgroups, instead of checking response as false, check for true. (Ref: AddTeamModal.tsx line number: 231)

### Relevant Feature or Code

_No response_

### Requirements (what)

_No response_

shubham-sp-07:
## Description
1. The group names start with /. The redirection on FE fails because the URL becomes // - issue has been fixed
Fixed an issue where group names starting with "/" caused front-end redirection to fail, resulting in URLs with double slashes (e.g., "//"). The URL handling has been corrected.
2. Fixed an issue on the team details page where the name and type were previously fetched from state, causing problems when accessing the page directly via bookmark or URL. Now, the team ID is obtained from the URL and the type is fetched from the backend endpoint, ensuring the page loads correctly in all scenarios.


anurag91jain:
@shubham-sp-07 need to update the PR. As discussed with Maher, there can be a group with the same name and different type. Hence it won't be possible from BE to decide based on the name which team is being opened.

Solution- when the user is selecting a team, the URL should be /<team-name>--<team-type>. The type should be fetched from the URL instead of making an API call

anurag91jain (approved):

PR #1783, opened 2025-08-28
Fixed the charts disappearing bug

### LINKED ISSUE #402: Only two charts are visible horizontally after adding more than two charts inside welcome block container

Gowrishankar-Palanisamy:

  --> Fixed the multiple charts disappearing bug .


johbaxter:
Can we create a setting for this?

johbaxter (changes requested):
Create setting for this

PR #1785, opened 2025-08-28
1764 add tags in app versions table

shubham-sp-07:
## Description
- Added AddTagModal component so users can add tags to specific commit IDs
- Extracted version management API functions to separate versions.ts file
- Consolidated duplicate pixel execution and error handling logic


shubham-sp-07:
image
image
image

PR #1729, opened 2025-08-19
Centralize CSRF fetching logic to avoid redundant calls and ensure consistent CSRF inclusion.

Vishal24-Kanini:
## Description
This PR refactors the SDK and client-side implementation to address an issue where CSRF tokens were not consistently sent with API calls, particularly on settings pages.

## What was happening before
- Each request independently attempted to fetch a CSRF token.
- This caused multiple redundant fetchCSRF calls and inconsistent behavior in some parts of the app.

## Changes Made
-  Centralized CSRF fetching logic after initial config is set.
- Ensured CSRF token is fetched once and reused for subsequent API calls.
- All outgoing requests are now handled via a custom interceptor, which automatically injects the CSRF token

image
image

PR #1821, opened 2025-09-04
Centralize CSRF fetching logic to avoid redundant calls and ensure consistent CSRF inclusion.

### LINKED ISSUE #1704: CSRF isn't being sent on all Settings pages

### Description

Image

- [x] All of our settings pages the endpoint does not get sent with csrf token
- [x] Refactor sdk and client implementation to only need the csrf once.

Image

### Relevant Feature or Code

_No response_

### Requirements (what)

_No response_

Vishal24-Kanini:
## Description
This PR refactors the SDK and client-side implementation to address an issue where CSRF tokens were not consistently sent with API calls, particularly on settings pages.

## What was happening before
- Each request independently attempted to fetch a CSRF token.
- This caused multiple redundant fetchCSRF calls and inconsistent behavior in some parts of the app.
- After logout the session, user faced issue with login because of CSRF not being reset.

## Changes Made
- Centralized CSRF fetching logic after initial config is set.
- Ensured CSRF token is fetched once and reused for subsequent API calls.
- All outgoing requests are now handled via a custom interceptor, which automatically injects the CSRF token.
- Reset the token after logout user can login easily after logout.

image

image


johbaxter:
https://github.com/SEMOSS/semoss-ui/pull/1729

I remember having issues with this PR, ensure this passes all test cases.  It gets sent with EVERY http request (settings pages were a problem, from what i remember)

Vishal24-Kanini:
Yes I had resolved the problem where token wasn't attaching with settings page earlier. In this PR I have ensured that It gets sent with EVERY http request. it has passed all the test cases and token is sent with every http request. also modify the code so token should reset after logout and user login again without any issue.

neelneelneel:
Should not be required with https://github.com/SEMOSS/semoss-ui/pull/2714

PR #1839, opened 2025-09-09
feat(renderer): fixed Overlapping issue for adding multiple charts

### LINKED ISSUE #401: Charts overlap each other when three different charts are added to the page

1. Navigate to the application.
2. Navigate to App and edit the created app.
3. Drag and drop an Area chart onto the page.
4. Drag and drop a Bar chart below the Area chart.
5. Drag and drop a second Bar chart below the first Bar chart.

### Expected Result

Each chart should be displayed in its own section, clearly separated from the others.

### Actual Results

The charts overlap each other, making it hard to distinguish and use each chart.

### Additional Context

Image

### Requirements or Tasks

_No response_

- [x] Yes

- [x] Yes

pallabi2303:
## Description
Fixed Overlapping issue for adding multiple charts

## How to Test
1. Navigate to the application.
2. Navigate to App and edit the created app.
3. Drag and drop an Area chart onto the page.
4. Drag and drop a Bar chart below the Area chart.
5. Drag and drop a second Bar chart below the first Bar chart.

Below video shows how it is working currently:
https://github.com/user-attachments/assets/34be0ece-aa91-4a62-b0a1-359f4432d908


libs/renderer/src/components/block-defaults/echart-visualization-block/VisualizationBlock.tsx
```diff
@@ -27,7 +27,7 @@ const StyledNoDataContainer = styled("div", {
 const StyledDataContainer = styled("div", {
 	shouldForwardProp: (prop) => prop !== "error",
 })<{ error?: boolean }>(({ error = false, theme }) => ({
-	minHeight: "50%",
+	minHeight: "450px",
```
johbaxter: What if its parent container block is 300px.  Did we test this?

johbaxter:
Please fix conflicts

libs/renderer/src/components/block-defaults/echart-visualization-block/VisualizationBlock.tsx
pallabi2303: I have changed the min-height to 350px, and below is the screenshot of how it looks now:
image

here the bar chart and pie chart are inside the container block with size 300px and the area chart is outside the container.

PR #1906, opened 2025-09-18
Calendar view block in drag and drop

aditiiisinhaa:
## Description
Creation of Calendarview Block in drag and drop

## Notes
Here is the screenshot:
image


johbaxter:
So i sorta like having a separate component/block for this.  But also not too sure.

Theoretically you should be able to attempt to build a calendar with our blocks that we have now.

But the styles may/not be a pain to get working.

If you setup a call with me to demo this i can give you feedback.

@aditiiisinhaa @anurag91jain

PR #1907, opened 2025-09-18
464 fix checklist functionality for visualization filter block

### LINKED ISSUE #464: Fix Checklist Functionality for Visualization Filter Block

1.) Create a new drag and drop app.
2.) Pull in data / frame via method of your choosing.
3.) Create visualization filter block with checklist selected as method of choice.
4.) Check (select) items to filter and click apply.
5.) You will see that an array of integers/NaN is being passed in as the values to filter on, and not the string values selected

### Expected Result

The checked (selected) items in the visualization filter block should filter based on the string values selected, and not attempt to convert those items into an array of integers.

### Actual Results

An array of integers is created from the checked values for filtering.

### Additional Context

Image

**We expect the value to be passed in to the filter to be ['A'], but what we are seeing is [NaN]**

### Requirements or Tasks

- [ ] Update the function being called on apply to not transfer the checklist values into an array of integers

- [x] Yes

- [x] Yes

tzylks:
## Description
Update modularity of VisualizationFilterBlock components. Fix bugs related to functionality of filtering. 

## Changes Made
- Increased modularity of components by separating the logic of dropdown / multiselect / slider / checklist into their own components.
- Addressed bugs related to filtering (prior, the function being run was trying to convert values into an array of integers for all component options -- this should only occur when using the slider, however)
- Addressed bug with multiselect component where chips were not being reset

## How to Test
1. Create drag and drop app > use preferred method of getting and setting data in dataframe > create Visualization Filter component > choose how you want to filter data (dropdown, multiselect, etc.) > select data to filter > click apply button and check the calls being made to ensure data is correctly filtered in the frame.
2. Use the multiselect component and close out any number of chips. Ensure chips are reset when clicking the unfilter button.


ehynd:
@trevorzylks @antgibson couple things I found while testing! 

1. If you unfilter using the unfilter button, but something has checkboxes (ie the checklist or the dropdown), the checkboxes should all be checked the next time you open the filer.
Ex: Create a dropdown filter, save the app, in view mode use the filter and select 1+ options, use the unfilter button on the filter.  When I open the dropdown again, all options should once again be checked since all options are selected due to the unfilter. This happens on the checklist as well, when I use the unfilter button, all the options should default to checked. 

2. Dropdown: while in edit mode, filter label doesn't update automatically when changing columns.
Ex: Create a dropdown filter, select a column, click update. Now within the same filter, select a different column and click update. The filter technically updates, but the label will still say "Filter of og column" until manually updated 

3. Multi select: how do you get the options back once you deselect them? 
Ex: I have options 1, 2, and 3. I accidentally unselect 1 and 2, but I really only wanted to unselect 1. How do I get 2 back without unfiltering everything and starting over? 

4. Multi select: if you try to filter with nothing selected (all options deleted), the filter reverts to all options, but the charts don't. 
Ex: I created a filter with male and female. I x'd out of male and clicked apply. Then I x'd out of female as well and hit apply. Both male and female popped back into the filter, but the charts stayed filtered to just female. 

5. Slider: My selection should stay after I hit apply. 
Ex: I had a slider for age and selected from 20-27 and hit apply. Everything populated correctly, BUT, the slider reverted to showing 20-67 (the full range).  It should show the slider dots on 20-27 and not have the entire slider highlighted.

tzylks:
@antgibson @ehynd additional updates have been made and pushed. I believe we have tackled all items from the checklist.

PR #1965, opened 2025-09-30
feat(client): applied same json/vitest configuration from the coverag…

boredcod:
…e branch

## Description

Modified json and vitest configuration to allow coverage tests
## How to Test

Go into any testing directory (ex: libs\renderer\src\testing\block-defaults\) and run: npm run test:coverage [name of the testing file]
3. Expected outcomes
Command line should show percentages of coverage

### DIFF
--- a/libs/renderer/package.json
+++ b/libs/renderer/package.json
@@ -17,7 +17,8 @@
         "build": "rollup -c --environment NODE_ENV:production",
         "build:dev": "rollup -c",
         "test": "vitest run",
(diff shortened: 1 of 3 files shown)


johbaxter:
@boredcod and @memisrose Can we put together a PR that moves all spec files associated to the renderer and ,ove them to their respective `<block-name>-block` folder

I know i went back and forth with you on this a while ago and gave you an incorrect approach last time Rose but i think this will be better for organization if we go with above

PR #1970, opened 2025-10-01
1957 remaining members not visible after deleting team settings

### LINKED ISSUE #1957: Remaining members not visible after deleting selected members until page refresh

1.Navigate to Team Permission on setting page 
2.Navigate to the Add Member section.
3.Add one or more members.
4.Search for members by a common name.
5.Select all members from the search results.
6.Delete the selected members.
7.Observe the members list.

### Expected Result

After deletion, the remaining members should be immediately visible in the members list without requiring a page refresh.

### Actual Results

The remaining members are not visible immediately. They are only displayed after refreshing the page.

### Additional Context

When a user adds one or more members, then searches by the same name, selects all, and deletes them, the remaining members do not appear in the list immediately. The members are only visible after manually refreshing the page.

### Requirements or Tasks

1.Member List

Image
2.Serch by common name and delete

Image

3.After Deleted

Image

4.After Refresh

Image

- [x] Yes

- [x] Yes

bannaarisamy-shanmugham-kanini:
## Description
This PR fixes when a user tries to search for a user and deletes the same user from list. After deleting, it shows no members found ui. yet there are other records available, out of the search keyword. So if no records matched, there should be no members found message in the table.

## Changes Made
Fixed user search based delete, and then showing no members found row, while search keyword doesn't matched with any records

## How to Test

1. Go to Settings -> Team Settings -> select a team and go into it.
2. Choose members from team by searching with a keyword.
3. check all the filtered team members and then click delete selected.
4. Confirm the popup and wait for the table. it should now show No members found.

### DIFF
--- a/packages/client/src/components/settings/UserTable.tsx
+++ b/packages/client/src/components/settings/UserTable.tsx
@@ -1,7 +1,6 @@
 import { Add, Delete, Edit } from "@mui/icons-material";
-import CopyAllIcon from "@mui/icons-material/CopyAll";
 import SearchIcon from "@mui/icons-material/Search";
-import { useEffect, useMemo, useRef, useState } from "react";
+import { useMemo, useRef, useState } from "react";
 import { useDebouncedValue } from "@semoss/sdk/react";
 import {
 	Avatar,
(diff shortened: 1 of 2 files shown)


johbaxter:
Members table could still use some work

1. Move to second page of Members table and search for a user that was on first page4
2. Add multiple members at the same time. Flashes multiple toast messages.

PR #1990, opened 2025-10-06
Fixed functionality to the Navigate onClick that allows the user to select the destination as page-1 

### LINKED ISSUE #1958: Can't navigate to page 1

### Description

Right now, there is an onClick action called "Navigate".  However, the user can't navigate to page 1 / the first page created. 

There is technically a little green space you can click above page--2, but it does not allow the user to save and use that to navigate to page 1. 

This is needed as users may want to be able to create multiple pages and navigate back and forth between them. 

### Relevant Feature or Code

Image

### Requirements (what)

- [x] Add functionality to the Navigate onClick that allows the user to select the destination as page 1 / the home page of the drag and drop app

Vishal24-Kanini:
## Description
In this PR i have fixed the issue related with the functionality to the Navigate onClick that allows the user to select the destination as page-1 

## Changes Made
- I have found page.route was empty for page-1, that's why user wasn't able to see and select the page-1 in the Navigate Onclick dropdown So changed it to page.id that works' fine.

## Fix Demo
https://github.com/user-attachments/assets/82ebcb82-b75b-47ad-a081-d973e1026e41

## How to Test
1. add onClick action called "Navigate".
2. see the dropdown use is able to select page-1/homepage and navigate to it


libs/renderer/src/store/state/state.store.ts
```diff
@@ -2058,7 +2058,7 @@ export class StateStore {
 					const pageBlocks = this.getAllBlocksOfType("page");
 
 					const urlPageRouteMatch = pageBlocks.find(
-						(page) => page.data.route === destination,
+						(page) => page.id === destination,
```
ehynd: can we actually do this differently and modify in PageBlockDef? 

route should be part of the PageBlockDef tsx under data. after adding that in, we should be able to pull it directly from there

rameshpaulraj: so i have made route as a part of PageBlockDef tsx under data. But the main issue i have found in this is when we creating an app , page-1 is created by default and route has been sent as empty string like route = "" inside json param of CreateAppFromBlocks API.

Even after added route under data in PageBlockDef this only sets the route at runtime but it does not update the underlying MobX state or the original block definition.
I have modify the route of page-1 from empty string to page-1  under BASE_PAGE_BLOCKS , then page-1 always contains it's route , @ehynd please take a look at these changes.

johbaxter (changes requested):
This is not the fix.  All you did was default the route to `/page-1`

Instead in the settings component for navigation events add a check in the code for routes that are an empty string.  If its an empty string show the display value as `/`

johbaxter:
@ehynd, @Vishal24-Kanini, @rameshpaulraj i can explain in a call if need be.  

This is not the fix.

**ATTENTION: DO NOT APPROVE**

Vishal24-Kanini:
Hi @johbaxter ,  with the earlier changes was trying to handle the issue at root level, didn't know the faith of default route to empty string, that was just a brute force approach.

Now I have modified it and handle the route for home page as **/** under settings component for navigation events.

There is also another change in function (dispatchOpenEvent) under state of page block , so there should be equality between data.route and destination when hit the button to navigate . 

please take a look at these changes.

johbaxter:
https://github.com/SEMOSS/semoss-ui/commit/ac42116da9363bfd80537181e2988c6bdcf2d14d

PR #2002, opened 2025-10-08
feat(client): make apps created by the current user easier to find

snehakumari369:
## Description
Added a new toggle button called 'Created By Me' in the 'Filter By' section. When clicked, it displays only the apps created by the currently logged in user.

## Changes Made
Added and updated logic in
- AppCatalogPage.tsx 
- Filterbox.tsx

## How to Test
1. Navigate to the 'Apps' section and click on the 'Created By Me' button in the 'Filter By' section.
2. Verify that only the apps created by your logged-in account are displayed.


johbaxter:
@snehakumari369 This is a FE solution, ideally would like this filter sent to the BE

anurag91jain:
@johbaxter Should the "created by me" filter option (left section) be generated from BE or should the results (right section) be generated from BE? We tried to avoid an additional call to fetch the results. We can add BE changes for the additional filter option.

snehakumari369:
Here are the screenshots for both the implementations:
1. Left Section ('Created By Me' button under 'Filter By' section)

image

2. Right Section ('Created By Me' section on top of 'All Apps' section - current implementation)

image

image

Srk15:
Closing this PR as the ticket needs a BE filter implementation. Changes for the same will be raised as part of a separate PR.

PR #2005, opened 2025-10-08
fix(libs/renderer): show placeholder text if data has not loaded

### LINKED ISSUE #497: Update Text Block to Include Ability to Add in Data from a Data Frame
Add ability for a text block to pull in data from a data frame.

tzylks:
## Description
Show placeholder text if variables have not been loaded in (e.g. making a notebook call)

## Changes Made
Add placeholder text if no variable has not been loaded

## How to Test
1.) Assign text block text value to available cell variable
2.) If the variable has not been set through running the cell, you should see "Waiting for value..."

## Notes
The original ticket requirements have changed for this PR. Only updating placeholder text until a better solution has been discussed and architected.


johbaxter:
Not a fan of this, when someone builds there app.  and it entails quite a bit of text blocks that are dependent on a variable.

You will have a bunch of portions on the app that say "Waiting for text"

Im fine with showing some indicator, but I'd like it to be optional for the user to have it displayed.

Maybe add a setting?

tzylks:
> Not a fan of this, when someone builds there app. and it entails quite a bit of text blocks that are dependent on a variable.
> 
> You will have a bunch of portions on the app that say "Waiting for text"
> 
> Im fine with showing some indicator, but I'd like it to be optional for the user to have it displayed.
> 
> Maybe add a setting?

Makes perfect sense to me. Added in the ability to enable the placeholder from the text block menu. If not enabled. No value will appear. If enabled, the user will see "Waiting for value..."

johbaxter (approved):

PR #2006, opened 2025-10-08
feat: add create agent modal and banner

travon-speller:
## Description

A new modal to create a workspace when the button on discover page is clicked.
## Changes Made

Created Modal to make a workspace 
## How to Test

2. Expected outcomes


packages/playground/src/components/agent/AgentModal.tsx
```diff
@@ -1,43 +1,444 @@
...
 	 */
 	const isCreatingNew = agentInfo === null;
 
+	const [isLoading, setIsLoading] = useState<boolean>(false);
+	const [selectedTools, setSelectedTools] = useState([]);
+	const [tools, setTools] = useState([]);
+
+	// pixel call to get all tools
+	const getApps = usePixel<(Engine | App)[]>(
+		`MyEngineProject (metaKeys = ["tag", "description"], metaFilters=[{"tag":["MCP"]}], type=["PROJECT", "STORAGE", "DATABASE", "FUNCTION"], filterWord=[""])`,
```
tevanburen, line 101: On this page we only want MCPs right? I think those are only projects

packages/playground/src/components/agent/AgentModal.tsx
```diff
@@ -1,43 +1,444 @@
...
+	useEffect(() => {
+		if (getApps.status !== "SUCCESS") {
+			setIsLoading(true);
+			return;
+		}
+
+		const tools = getApps.data.map((tool) => getTool(tool));
+		setTools(tools);
+		setIsLoading(false);
+	}, [getApps.status]);
```
tevanburen: i believe this should also watch for getApps.data

packages/playground/src/components/agent/AgentModal.tsx
```diff
@@ -1,43 +1,444 @@
...
+			AGENT_CONTEXT: "",
+			AGENT_TOOLS: null,
+		},
+	});
+
+	const watchAll = watch();
+
+	const isFormValid = useMemo(() => {
+		return !!getValues("AGENT_NAME");
+	}, [watchAll]);
```
tevanburen: I believe it would be better to leave this as a constant rather than memoizing

packages/playground/src/components/agent/AgentModal.tsx
```diff
@@ -1,43 +1,444 @@
...
+									);
+								}}
+							/>
+							<Controller
+								name={"AGENT_TOOLS"}
+								control={control}
+								rules={{}}
+								render={({ field }) => {
+									return (
+										<Select
```
tevanburen: If we use the Autocomplete component here we can make this field a lot more simple

packages/playground/src/components/agent/AgentModal.tsx
tevanburen, line 1: A couple small differences in spacing on the figma

https://www.figma.com/design/7c21rDM4S4xHQYILPRpusC/Playground?m=auto&node-id=4054-25620

packages/playground/src/stores/chat/chat.store.ts
```diff
@@ -263,6 +263,48 @@ export class ChatStore {
 		}
 	};
 
+	addWorkspace = async (data): Promise<string> => {
+		try {
+			const esc = (v: string) =>
```
tevanburen, line 268: You should be able to just JSON.stringify all of the strings to do the same thing, no?

packages/playground/src/components/agent/AgentModal.tsx
travon-speller: I just copied this exact call from the one we use in the settings when we open the add tool modal.

packages/playground/src/stores/chat/chat.store.ts
travon-speller: Did it this way based on the suggestion in the PR Suggestion by Codium since it was a medium security risk.

packages/playground/src/components/agent/AgentModal.tsx
travon-speller: updated

packages/playground/src/components/agent/AgentModal.tsx
travon-speller: updated. Was originally using the stacked Select like the mocks said but since thats deleted autocomplete has all the features i manually added. good catch

packages/playground/src/components/agent/AgentModal.tsx
travon-speller: updated

packages/playground/src/components/agent/AgentModal.tsx
travon-speller: updated

antgibson96 (approved):

PR #2011, opened 2025-10-09
feat(renderer): modal block close event fix

### LINKED ISSUE #1941: Modal block close icon and outside click close event is not working as expected

1. Add a modal block to the workspace
2. Attach a dynamic boolean variable to the modal open state
3. Add a button block to the workspace and attach a onClick event listener to it, to make the modal open by switching the same variable state.
4. Now in the interactive/end-user mode onClick of button block open the modal block
5. Now try to close the modal using the cross icon(close) or outside click the modal.
6. Retry to open the modal again.
7. Now you notice the modal block will not open as expected.

### Expected Result

Modal should open and close seamless on proper settings

### Actual Results

Modal is not opening on the second instance if it got close using the cross/outside click in the previous instance.

### Additional Context

_No response_

### Requirements or Tasks

_No response_

- [x] Yes

- [x] Yes

### LINKED ISSUE #1968: QA| Modal block close icon and outside click close event is not working as expected

### Feature

Issue 1:

Steps to reproduce:

1. Log in to AICore.
2. Navigate to the “Apps” section.
3. Create a new app.
4. Add a Modal Block to your app.
5. Do not configure the Modal Block  only change open Modal to true, leave default settings.
6. Go to the breadcrumb navigation menu.
7. Observe the behavior or outcome.

Actual result:- Modal block remains open, preventing access to edit screen and all background options.
Expected Result:- User can access the edit screen and background options as intended.

https://github.com/user-attachments/assets/60205f9c-af28-4d01-b57b-8c044ebdbedd

### Component / Function

_No response_

### Given

_No response_

### When

_No response_

### Then

_No response_

Paulson-Robert:
## Description
Couldn't able to close/navigate anywhere in the app if the modal if not configured properly

## Changes Made
Made the modal backdrop restrict only to the specific app

## How to Test
1. Add a modal block to the workspace
2. Save the app without configuring the modal block
3. Go to interactive mode of the app
4. Now you can notice that the modal is default open state and not able to close or navigate anywhere as the modal backdrop covers all the options.
5. Fix : Now the backdrop is set to the parent(app) container, so now we can access the top nav bar though the modal is in open state.

## Notes
https://github.com/SEMOSS/semoss-ui/issues/1941#issuecomment-3384311686


libs/renderer/src/components/block-defaults/modal-block/ModalBlock.tsx
```diff
@@ -235,14 +236,20 @@ export const ModalBlock: BlockComponent = observer(({ id }) => {
 				<MuiModal
 					open={shouldShowModal}
 					onClose={handleClose}
-					// TODO: Switch to parent page
-					container={() => document.getElementById("page-1")}
+					container={() => {
+						const elements = Array.from(document.querySelectorAll('[id="page-1"]'));
```
johbaxter, line 240: Hey can we fix this to go off of the active page, not off of `page-1`.

johbaxter (approved):
The modal on close is still tied to page-1, but will address later i guess

PR #2017, opened 2025-10-10
Keep workspaces in store

tevanburen:
## Description
- Store workspaces in the chat.store, rather than fetching them on every page that needs them

## Changes Made
- Misc styling fixes relating to Neel's latest branch
- Store agents in chat.store
- When calling for an agent, just grab from chat.store
- Fix in new room page, needed to make the init function async to catch errors correctly
- In chat.initialize, grab all the workspaces
- In room.store, add a function to call UpdateRoomOptions. call that when initializing a room
- When editing a room's configuration, on deletion and addition for tools, call updateRoomOptions
- Don't pass mcpToolID to ask, because it will be pulled dynamically from the roomOptions
- Rename "agent" to "workspace"
- Add getTools to chat.store
- Fix typing in WorkspaceModal

## Notes
- If a user loses access to a tool in a room, they will see issues


tevanburen:
Notes to Neel:
- room.store is set up so that the first runRoomPixel will create a new insight - so we don't need to worry about explicitly initializing new rooms, because their first call will set insightId
- I did make sure to combine the init pixels

antgibson96 (approved):

PR #2056, opened 2025-10-22
fix(client): editing a job now works and uses the correct components …

travon-speller:
…and time

## Description

Editing a job always failed and it used a default cronExpression rather than the one for the job. Custom cron expression only accepted numbers or an asterisk.
## Changes Made

Removed the extra asterisk when clicking save for editing which was causing the cronExpression to be invalid.
Pass in the cronExpression instead of the whole builder to show the correct times.
Added extra validations to show the custom job when the cronExpression doesn't have numbers for minutes or hours.
Created a new custom cron validator to accept all legitimate cron inputs.
## How to Test

1. Go to the jobs page
2. Click edit on a job
3. Should see the time set as the jobs time.
4. should be able to make changes and click save and the changes persist
5. should be able to add custom cron input with slashes, commas, and dashes


packages/client/src/pages/jobs/JobCustomFrequencyBuilder.tsx
```diff
@@ -1,12 +1,11 @@
 import { useEffect, useState } from "react";
 import { Stack, TextField } from "@semoss/ui";
-import type { JobBuilder } from "./job.types";
```
antgibson96, line 3: why are we moving from job builder to just a string? do we lose anything here?

travon-speller: we were only using the cronExpression in the builder so just simplified it to just pass in the cronExpression instead of the whole builder object.

antgibson96 (approved):

PR #2075, opened 2025-10-27
1963 continue model catalog refactor

### LINKED ISSUE #1963: Continue Model Catalog Refactor
With Catalog import refactor, we will start with Model:
 
1. Match ModelImportForm.tsx component to designs
2. Fix Submission on ModelImportForm.tsx
3. Add necessary validations for Model form
4. Get Zip Upload working with Model import
5. Ensure all pre-existing model tiles are working in new component flow.  (Do not change data structures)

 
https://www.figma.com/design/E7x1Cpz6YZSCDHCJ6E3HQ6/Add-Model-Flow?node-id=0-1&p=f&m=dev

### Files associated with feature:

Image

### Zoom Recording regarding ask

Please ask @ehynd or @Paulson-Robert for the recording of this feature

JaganSomannaKanini:
## Description
This PR is regarding the refactoring of the model catalog.

## Changes Made

- Matching the ModelImportForm.tsx component to designs.
- Fixed Submission on ModelImportForm.tsx.
- Added necessary validations for Model form.
- Zip Upload working with Model import.
- Also added a new property in the data structure called "category" to specify the individual field to which category it belongs, this is helpful when we display the fields in the frontend.
- The description part for the General, Settings and the Credentials are also dynamic with respect to the model type. The constant for the same is being added to the data structure with the name "Category" alongside with providers.
- The "Custom" model for every type of models is being renamed with "Others", and the image is updated with the relevant model image.

## How to Test

- Go to Models catalog and click on Add Model.
- Select the type model in which you want to create the model.
- Fill out the required fields and hit create.
- The model should be created.

## Notes
image

image

image

image

image

image

image

image

image


johbaxter:
### OpenAI-Compatible:

- [X] Can we add a field called “Deployment Type” that will go into the init script as deployment_type. This should have options for “vLLM” and “TGI”.

**For OpenAI-Compatible, should it be the same init script as OpenAI? Assuming yes**

---

### OpenAI:

- [X] For the chat-type options, we need to add an additional option for “responses”

---

### OpenAI Azure:

- [X] I think that Deployment Name needs to be replaced with a field called API Version which should go into the init script as api_version and should be required.
- [X] We should also add “responses” to the chat-type options as we did above for OpenAI.

---

### Google Vertex:

- [x] I need a way to pass a Project field to the init script as project

**I added a project field to form, and also put placeholder for project in init script**

- [ ] I also need an optional field called Service Account Credentials that can be passed to the init script as service_account_credentials. This should be a larger text box and needs to maintain the formatting that is pasted in (json)

**added optional field, and added to init script**

---

### All Providers:

- [x] The selections for “Record Questions and Responses” and “Keep Conversation History” should always default to true in the dropdowns.

- [x] The AWS Bedrock Catalog seems to ask for Context Window but none of the other catalogs do. We should ask for that in each catalog and pass it to the init script as context_window

**I also added VAR_NAME and INIT_SCRIPT to aws bedrock, it was never there**

- [x] Is it possible to update the init script if the user changes a field? For example, if I edit the max completion tokens field, it does not update the init script? Therefore this field would not actually get used.

**I think your saying actually fill out the placeholders in the init script, and send to backend with filled out params as such**

"""import genai_client;johns_var = genai_client.OpenAiClient(model_name = 'gpt3.5', api_key = '3873787823', chat_type = 'responses)"""

**Also should i just disable that init script field, and it only gets updated when respective user updates field**

**Also is every field supposed to just be put into the init script. Stuff like MAX_INPUT_TOKENS and MAX_TOKENS**  


- [ ] Most of the embedding model selections for the different providers seems broken. If I click on the embeddings selection for OpenAI it brings up the old Davinci model.

**I'll have to touchbase with you on this**


- [x] I don't think we should even make users write in their own variable name, its kind of confusing and never really gets used by anyone but us when we are debugging. We could probably just make it default to something like openAiModel or whatever kind of model they are making.

**I can hide this field, sounds like its pointless to user**

johbaxter:
**A:**

image

Do we want this  ^

**B:**

image

Or this ^

johbaxter:
@JaganSomannaKanini Going to ask you to continue on my branch

This was a remark brought up in above comment

**_"Most of the embedding model selections for the different providers seems broken. If I click on the embeddings selection for OpenAI it brings up the old Davinci model."_**

---

In the data structure that we have in `model-import.constants`  

**below is where im referencing in the screenshot

image

You'll notice that we have 2 types of forms, one for regular llms

image
 
and one for embeddings

image

Currently this is our model catalog

image

---

What i need done is for you to make sure that when you click a tile that is labeled as embeddings

image

I need you to show these fields in the data structure

image

---

### Task


- [x] any tile that is embeddings in the catalog ensure that you show the respective form
- [x] Fix sizes of textField and Select boxes to match in size

image

johbaxter:
Would be nice if we beef up the catalog and show more tiles.

I put the link to various provider catalogs.  Important part is making sure the model id is correct.

image

johbaxter:
image

@ehynd @kunal0137 @ryanweiler92 

Alright so im merging this into dev, it is not done. (You can still upload zip no prob)

But it has the refactored code, and everything a bit easier to work on.

Im gonna focus on feedback with OpenAI first and then after we will fix the subsequent providers one by one ( should be easy once i get OpenAi reqs down )

---
**What i think is left over**
---

- Fix the init scripts to pull all the respective fields like MAX_TOKENS and such
- maybe change embeddings form fields and actually tag the model with 'embeddings'

PR #2193, opened 2025-11-11
Vitest for playground

memisrose:
## Description

- Upgraded to vitest v4
- Each package (that has a vitest.config) should be able to use pnpm run test:ui to launch the vitest browser
- Add vitest to playground
- run `turbo run test` at the root to run all tests

Blocker:
- unable to use turbo and run staged tests


neelneelneel (dismissed):
- [x] Add Lock File
- [x] Can we put tests next to the component itself? We can just have it end in .test.tsx

neelneelneel (approved):

PR #2197, opened 2025-11-12
feat(client): fixed issues on Team Permission page

### LINKED ISSUE #2189: Issues of team permission page

Team Permission Page is giving issue during creating team and add filter section

### Expected Result

_No response_

### Actual Results

_No response_

### Additional Context

_No response_

### Requirements or Tasks

_No response_

- [x] Yes

- [x] Yes

snehakumari369:
## Description
- When we click on add member no multiple calls are going.
- Add new team no longer throws an error on success.
- Showing user id, email, type in add members for teams similar to add members on app.
- Gave padding between members when we select multiple members to add.
- Fixed the issue where navigating to the next page and returning caused previously selected members (with same name) to appear duplicated.
- Removed the duplicate 'Added Date' column and fixed the column header positions.
- Added proper 'No Apps found.' and 'No Engines found.' messages when search results are empty, aligning the behavior with the members list.
- Resolved the styling bug where selecting a single member turned the email into a clickable link; it now remains non-clickable as intended.

## Changes Made
Added and updated logic in:
- AddTeamModal.tsx
- TeamMembersTable.tsx
- TeamEnginesTable.tsx
- TeamProjectsTable.tsx

## How to Test
- Create a new team - it should not throw any error.
- When we click on Add Members, member list should display both name and member id (email).
- When we search an app or engine and the search results are empty, it should throw proper error messages.
- When we navigate back from the next page, the selected members should not be duplicated.
- When we select multiple members to add, there should be no styling issues.


snehakumari369:
Issue: Add new team is throwing on error on success
Before:
image

After:
image

Issue: No padding/spacing or other stylying issue when we select multiple members to add and turned email into non-clickable link.
Before:
image

After:
image

Issue: When we want to Add members, option only shows Name. (Now it shows, Name, User Id, Email and Type)
Before:
image

After:
image

Issue: There were two columns for 'Added Date' for members, and the position of column headers was incorrect.
Before:
image

After:
image

Issue: Duplicate members appear when navigating back to previous page of members.
Before:
image

After:
image

Issue: When searching for Apps and Engines, if the search result is empty, it redirects to Add Apps/ Engines page, rather then giving a not found message.
Before:
image
image

After:
image
image

PR #2215, opened 2025-11-18
Resolved name param issue from app zip upload & app clone

### LINKED ISSUE #2173: Remove name param from app zip upload

### Description

When uploading an app zip, the name param is asked for in the modal. However, this is not ever passed to the app as it takes the name from the uploaded zip.  This field can be removed from the upload modal so the user does not get confused. 

### Relevant Feature or Code

Image

### Requirements (what)

- [x] Remove name field from upload app modal

Vishal24-Kanini:
## Description
In this PR i have fixed the issue related to name param between app zip upload & app clone. According to the requirement name param should be there while app cloning but not required while app zip upload.
I rendered the name param conditionally in both functionalities as they using the same component for name param.

## Changes Made
- conditionally render name param from AppDetailsStep component.
### name param will be there while cloning app.
image 

### name param will not be there while app zip upload
image

## How to Test

2. Expected outcomes


Paulson-Robert (approved):
Approved

PR #2230, opened 2025-11-19
MCP SDK changes

tevanburen:
## Description
- Pass roomId from playground to the child apps
- SetRoomForInsight on playground and child apps (a flag can disable this behavior)
- Expose tool info to child apps


libs/sdk/src/js-frameworks/react/contexts/Insight.context.tsx
```diff
@@ -78,15 +85,15 @@ export const InsightProvider = (props: InsightProviderProps) => {
...
-	}, [insight, insight.actions]);
+	}, [insight, insight.actions, syncInsight]);
 
 	// initialize the insight / destroy
 	useEffect(() => {
 		// initialize the insight
-		insight.initialize(options).finally(() => {
-			// update the state
+		(async () => {
+			await insight.initialize(options);
```
neelneelneel: This shouldn't be changed - .finally allows it to be run even if an error it thrown

packages/playground/src/pages/room-page.tsx
```diff
@@ -88,17 +89,15 @@ export const RoomPage = observer(() => {
...
-				};
+				tool: MCPToolResponse;
 			}>,
 		) => {
 			try {
-				if (!event.data || event.data.type !== "SMSS_EXEC_TOOL") {
+				if (
+					!event.data ||
+					event.data.type !== "SMSS_EXEC_TOOL" ||
+					room.roomId !== event.data.tool.roomId
```
neelneelneel: This isn't necessary. Will brake older apps calling tools.

packages/playground/src/stores/room/room.store.ts
```diff
@@ -358,7 +358,7 @@ export class RoomStore {
 					{ OPTIONS?: RoomStoreInterface["options"] }, // partial because this doesn't work for old rooms
 				]
 			>(
-				`GetPlaygroundMessages(roomId=["${this._store.roomId}"]); GetRoomOptions(roomId=${JSON.stringify(this._store.roomId)});`,
+				`GetPlaygroundMessages(roomId=["${this._store.roomId}"]); GetRoomOptions(roomId=${JSON.stringify(this._store.roomId)}); SetRoomForInsight(roomId=${JSON.stringify(this._store.roomId)});`,
 			);
```
neelneelneel, line 362: I dont think this needs to be called. @kunal0137 - is this accurate?

neelneelneel (changes requested):

packages/playground/src/stores/room/room.store.ts
kzsb03: Think it does, else how are we linking the room folder?

kzsb03 (approved):

kzsb03 (approved):

PR #2281, opened 2025-11-21
[DRAFT] 2211 add ability to batch database calls from notebook cells datagrid block

### LINKED ISSUE #2211: Add Ability to Batch Database Calls from Notebook Cells / DataGrid Block

### Description

As of right now, there is no way to batch calls when necessary (e.g. making a DB call that will return thousands of rows). We need a way to batch the DB calls from a custom SQL cell, and a FE solution for running the preceding/proceeding batch.

### Relevant Feature or Code

1.) Add ability to batch calls for the custom SQL cell
2.) Add ability to run batched calls via the DataGrid block

### Requirements (what)

1.) Add ability to batch calls for the custom SQL cell
2.) Add ability to run batched calls via the DataGrid block

tzylks:
## Description
Add in ability to batch calls via QueryImportCell, and call in additional batches via the data grid. The data grid UI will change if batching is enabled for the block associated with the frame. User will be able to filter, and has the option to enable exporting via the GridBlock settings. If enabled, an export button will appear that will allow the user to enable all fields that are present in the data grid. 

## Changes Made
1.) Add ability to batch calls via QueryImportCell
2.) Update UI for DataGrid when batching is enabled for the frame selected
3.) Add ability to enable exporting of DataGrid data
4.) Add ability to filter on DataGrid

## How to Test
1.) Create a custom sql cell.
2.) Enable batching and set the batch amount.
3.) Run the query
4.) Attach DataGrid block to frame.
5.) Go to block settings for DataGrid > tools > Enable Export for Data Grid
6.) Export Button should appear above data grid and export data via CSV when clicked
7.) Click on Filter button above data and ensure you can filter columns


ehynd:
two things! @trevorzylks 

image

is there a reason I can't delete the entire number in the box? I'm unable to change this to "40" in an easy way for example since I can't backspace on the "5" that's in the box.  

Could we add the batching to the bottom cell as well? This is the import data type from data catalog cell.

ehynd:
could you also please align all the tools to the left with icons on the right? this was an existing style issue but would be great to fix: 
current state: 
image


what it should look like: 
image

tzylks:
@ehynd pushed up some changes. Can you test again?

PR #2291, opened 2025-11-24
fix: add default style to overwrite tailwind

### LINKED ISSUE #2224: Text and heading blocks all show up the same

1. Go to drag and drop app 
2. Drag in text block 
3. Drag in Heading 1 block 
4. Notice that they show up the same 
5. This is true for all Heading blocks 

### Expected Result

Different heading options will be bolded/different sizes 

### Actual Results

Image

All text/heading blocks show up the same 

### Additional Context

_No response_

### Requirements or Tasks

- [ ] For simplicity sake, hide the 6 heading boxes. The same outcome can be achieved by just using the text block, and users won't be confused on the differences. 

- [x] Yes

- [x] Yes

travon-speller:
## Description

Updated global CSS to reset the base header tags styling to user agent

## Changes Made

Updated global CSS to reset the base header tags styling to user agent

## How to Test

1. Create/Go to a Drag and Drop app
2. Drag the different headers onto the canvas and see they default to the correct size

## Notes

Screenshot 2025-11-21 154718


neelneelneel (changes requested):
Updating the global styles will reflect in other places. I recommend updating it at the block level.

travon-speller:
updated it to use block level settings

neelneelneel (approved):

PR #2303, opened 2025-11-25
2216 vector catalog refactor

JaganSomannaKanini:
## Description
In this PR the refactoring of the vector database catalog is done.

## Changes Made

- The data structure of the vector database catalog
- File upload button is moved along side the search.
- The UI of the tiles is now matching the model catalog.
- The form UI has a new look based on the figma.

## How to Test

1. Select the Vector catalog.
2. You will see different types of database to add, select any one.
3. Fill out the form.
4. On submission the database is added to vector catalog.

## Notes
@johbaxter 

- In the data structure, for the "select" component, the options section has both display and value. The reason for keeping this is in some cases we have completely different value and display names.
- A new property is added "optionRule" to handle the pixel call for the select component.
- In the tiles I have retained the image instead of the name. If we do not provide the image the trimmed name would appear  same as in model catalog tiles.
- I have added the link also pointing to the official documentation of the respective database.

image

image
image


JaganSomannaKanini:
Duplicate PR

PR #2330, opened 2025-12-01
Function catalog refactor

JaganSomannaKanini:
## Description
In this PR the refactoring of the function catalog is done.

## Changes Made

- The data structure of the function database catalog.
- File upload button is moved along side the search.
- The UI of the tiles is now matching the model catalog.
- The form UI has a new look based on the figma.

## How to Test

- Select the Function catalog.
- You will see different types of functions to add, select any one.
- Fill out the form.
- On submission the function is added to function catalog.

## Notes

- In the tiles I have retained the image instead of the name. If we do not provide the image the trimmed name would appear same as in model catalog tiles.
- I have added the link also pointing to the official documentation of the respective functions.

image
image
image


JaganSomannaKanini:
Duplicate PR

PR #2332, opened 2025-12-01
534 Export Data Tool for Data Frames 

### LINKED ISSUE #534: Export Data Tool for Data Frames

### Problem

Drag & Drop apps don't have an automated way to create export buttons for data frames. Currently you have to manually make a notebook, create the pixel statements, create a button block and then attach the notebook to the cell. Whilst in the legacy app, the tool did this for users.  

### Solution

Create a tool that will take in: 

1. Drop down to select a valid data frame.
2. A way to select / filter columns for export 
3. Drop down to select download type for export.

### Alternatives

I have built out this tool as a Data Grid tool option, so we can keep the export tied to the Grid block. 
We can also, add this tool as a default Tool for blocks if they are tied to data frames (e-charts, Grid, etc)

### Additional Context

Image

Image

### Requirements or Tasks

- [ ] Decide where we want the tool to exist in the Drag & Drop builder application 
- [ ] Decide which reactors we want to use for the Exports Types.

JoshRome26:
## Description
We add in a new standalone Export Data panel that allows users to create export buttons or directly export data from data frames. We also add a new Export button for the flex layout on the right panel.

## Changes Made
- Toggle between creating an export button or directly exporting data
- Dropdown to select from available frames with manual refresh capability
- Select/deselect specific columns with a reset button to restore all columns
- Dropdown for multiple export formats (CSV, TSV, Excel, and Text) 

- Automatic notebooks creation with format [export_[frame]_[type]. As well, detection to reuses existing notebooks with identical export configurations to prevent duplicates.
- New Migration file that reloads the state of the layout to add Export button to older applications and to new applications.

## How to Test
1. Create a Drag & Drop Application.
2. Open the 'Export Data' panel by clicking the 'Export' button. Check that there is a warning of no valid data frames.
3. Create a notebook and create a Frame with valid data.
4. Click on the refresh button to update the dropdown and select the frame you created.
5. Select / Deselect columns from the frame. (Check if the reset button works)
6. Select Export Type
7. Create export button, and check if notebook and button was created.
8. Switch the toggle, and run the export and see if you get a valid export.
9. Go back to creating a button OR exporting live, and change up the different inputs and see if you get the desired outcomes.

### DIFF
--- a/libs/renderer/src/store/state/migration/MigrationManager.ts
+++ b/libs/renderer/src/store/state/migration/MigrationManager.ts
@@ -15,6 +15,7 @@ import migrate__1_0_0_alpha_13__to_1_0_0_alpha_14 from "./migrate__1_0_0_alpha_1
 import migrate__1_0_0_alpha_14__to_1_0_0_alpha_15 from "./migrate__1_0_0_alpha_14__to___1_0_0_alpha_15";
 import migrate__1_0_0_alpha_15__to_1_0_0_alpha_16 from "./migrate__1_0_0_alpha_15__to___1_0_0_alpha_16";
 import migrate__1_0_0_alpha_16__to_1_0_0_alpha_17 from "./migrate__1_0_0_alpha_16__to___1_0_0_alpha_17";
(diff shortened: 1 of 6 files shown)


johbaxter:
Im not a fan of where we put this, but i'll push this for the sake of making it easier to create exports for no code users and getting a net new feature in

PR #2376, opened 2025-12-08
Resolve 2375 Successful filter feedback for visualization filter block 

JoshRome26:
## Description
Add in successful alerts that will appear on the top right of the application when a filter has been applied or removed by the user. 

## Changes Made
Added success notifications on the handleReset & handleApply functions that will appear after successful pixel calls in the mentioned functions. 

## How to Test
1. Create a Drag & Drop application.
2. Create a notebook and run a Data Import Cell with test data.
3. In the block view, add a Visualization filter and data grid component. Attach the filter and grid with the data.
4. Apply Filters and see if the new notification appears on a successful Apply.
5. Remove filters and see if the new notification appears on a successful Reset.

### DIFF
--- a/libs/renderer/src/components/block-defaults/visualization-filter-block/ModularFilterComponents/FilterChecklistComponent.tsx
+++ b/libs/renderer/src/components/block-defaults/visualization-filter-block/ModularFilterComponents/FilterChecklistComponent.tsx
@@ -1,5 +1,5 @@
-import { Box, Button } from "@mui/material";
 import { useEffect, useState } from "react";
+import { Box, Button } from "@semoss/ui";
 import type { FilterComponentProps } from "../filter";
 import FilterIconComponent from "./FilterIconComponent";
 import FilterListComponent from "./FilterListComponent";
@@ -29,7 +29,7 @@ const FilterChecklistComponent: React.FC<FilterComponentProps> = ({
 		}
 	}, [resetKey, mode, JSON.stringify(checkedValues)]);
 
-	const handleApply = async () => {
+	const handleApply = () => {
 		let selected: string[] = [];
 
 		selected = checked;
--- a/libs/renderer/src/components/block-defaults/visualization-filter-block/VisualizationFilterBlock.tsx
+++ b/libs/renderer/src/components/block-defaults/visualization-filter-block/VisualizationFilterBlock.tsx
@@ -1,6 +1,12 @@
 import { observer } from "mobx-react-lite";
 import { type CSSProperties, useEffect, useState } from "react";
-import { Box, styled, Typography, useNotification } from "@semoss/ui";
+import {
+	Box,
+	CircularProgress,
+	styled,
+	Typography,
+	useNotification,
+} from "@semoss/ui";
 import { useBlock, useBlocks } from "../../../hooks";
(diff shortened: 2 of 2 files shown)


libs/renderer/src/components/block-defaults/visualization-filter-block/ModularFilterComponents/FilterChecklistComponent.tsx
```diff
@@ -1,5 +1,7 @@
-import { Box, Button } from "@mui/material";
+// import { Box, Button } from "@mui/material";
```
antgibson96: remove

libs/renderer/src/components/block-defaults/visualization-filter-block/ModularFilterComponents/FilterDropdownComponent.tsx
```diff
@@ -44,7 +45,10 @@ const FilterDropdownComponent: React.FC<FilterComponentProps> = ({
 		if (mode === "dropdown") {
 			selected = checked;
 		}
-		onApply(selected, mode);
+		setIsLoading(true);
```
antgibson96: Seems like we are duplicating this logic, can we consolidate the loading somehow?

JoshRome26: I added in a hook that will be used to do the loading state and the on Apply functionality, instead of having the duplicated logic.

libs/renderer/src/components/block-defaults/visualization-filter-block/VisualizationFilterBlock.tsx
```diff
@@ -90,63 +91,74 @@ export const VisualizationFilterBlock: BlockComponent = observer(({ id }) => {
...
-		if (mode !== "slider" && mode !== "multiselect") {
-			setData("selectedValues", selected);
-		}
 
+		setIsLoading(true);
 		try {
-			for (let i = 0; i < data.frame.length; i++) {
-				// Construct the command to set a filter on the frame based on the selected column and values
-				const pixelCommand = `META | UnfilterFrame(${data.frame[i]});`;
+			//set initial value of valuesString -- will change based off mode and array length
```
antgibson96: why are we rewriting all of this?

libs/renderer/src/components/block-defaults/visualization-filter-block/VisualizationFilterBlock.tsx
```diff
@@ -224,6 +246,7 @@ export const VisualizationFilterBlock: BlockComponent = observer(({ id }) => {
 							onReset={handleReset}
 							color={data.color}
 							size={data.size}
+							isLoading={isLoading}
```
antgibson96: Can we just wrap this entire if statement in a loading component?

libs/renderer/src/components/block-defaults/visualization-filter-block/VisualizationFilterBlock.tsx
JoshRome26: I added a try statement that handles the loading state for the logic, so most of it is just indentation to accommodate that.

PR #2430, opened 2025-12-15
Timestamp column Query output table

### LINKED ISSUE #2307: Timestamp column in the query output table displays time values with more digits after the seconds

1. Open the application
2. Create Database
3. Create drag and drop application
4. Create Notebook
5. Import data via 'Custom Import(SQL)'
6. Execute the cell
7. Mouse hover the below hidden options
8. Select 'Transformation' hidden option
9. Select 'Timestamp' function
10. Select the frame, enter column name and select Include time checkbox
11. Execute the cell
12. See the timestamp column values

### Expected Result

Timestamp values should display up to seconds or, if required, limited fractional seconds

### Actual Results

Timestamp values include excessive digits after seconds, reflecting nanosecond precision.
Example: 2025-11-24 15:45:12.123456

### Additional Context

Image

### Requirements or Tasks

_No response_

- [x] Yes

- [x] Yes

NaveenRamasamy028:
Now user will be able to see upto seconds in the Timestamp

### DIFF
--- a/packages/client/src/components/notebook/operations/FrameOperation.tsx
+++ b/packages/client/src/components/notebook/operations/FrameOperation.tsx
@@ -76,6 +76,12 @@ export const FrameOperation = observer((props: FrameOperationProps) => {
 
 	const parent = tableRef?.current?.parentElement;
 	const parentParent = parent?.parentElement as HTMLElement | null;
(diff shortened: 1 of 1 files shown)


Paulson-Robert (approved):
Approved

PR #2434, opened 2025-12-17
Expand/collapse button state sync up with manual expand/collapse icons

### LINKED ISSUE #557: Collapse/Expand All button does not update based on manual expand/collapse actions for tables

1. Navigate to database catalog 
2. clicks on add database button and upload a database
3. click on query tab
4. Collapse all tables using either the caret or the "Collapse All" button.
5. Observe that the button changes to "Expand All".
6. Manually expand the SCHOOLS table by clicking its caret.
7. Observe that the button still reads "Collapse All" instead of updating to reflect the new state.

### Expected Result

When all tables are collapsed, the button should display "Expand All".
When all (or any) tables are expanded, the button should display "Collapse All".
The button label should always reflect the true state of the sidebar, regardless of whether it's changed via the button or manual (individual) table actions.

### Actual Results

The button text can become out of sync with the expand/collapse state if actions are performed manually rather than through the button.
This could lead to confusion about what action the button will perform.

### Additional Context

Image

### Requirements or Tasks

_No response_

- [x] Yes

- [x] Yes

Vishal24-Kanini:
## Description
Fixed the issue related to synchronization of states between expand/collapse button and individual tables.

## Changes Made
- Derive a live value from the table instead of stale state.
- Now internal expand Tables is updated synchronously inside the same state update along with toggle state.
- Also make a minor change in SQLQueryEditor.tsx so the editor border should not collides with outer box

image

image

## How to Test

2. Expected outcomes


packages/client/src/components/database/DatabaseStructureBrowser.tsx
```diff
@@ -149,6 +149,9 @@ export const DatabaseStructureBrowser: React.FC<
 	onGenerateQuery,
 	generateSelectedColumnsQuery,
 }) => {
+	// trying to put "Expand All/Collapse All" button in sync with expand/collapse icons
+	const allExpanded = searchedStructure.length > 0 && searchedStructure.every((t) => !!expandedTables[t.table]);
```
Paulson-Robert: Add proper null safety and remove consoles

Paulson-Robert (changes requested):

packages/client/src/components/database/DatabaseStructureBrowser.tsx
Vishal24-Kanini: Yeah sure, will do that

Vishal24-Kanini: removed consoles , you can check

Vishal24-Kanini:
@Paulson-Robert , I have re-fix the issue in the same branch and pushed changes. pls check.

Paulson-Robert (approved):
approved!

PR #2451, opened 2025-12-18
feat(client): error handling for json file types

### LINKED ISSUE #2435: JSON Validation

### Description

Currently, when an invalid JSON file is uploaded, the system throws an error and blocks the upload. Instead, the upload should be allowed, and any JSON validation errors should be displayed within the file editor.

### Relevant Feature or Code

_No response_

### Requirements (what)

_No response_

KirthikaKumar-K:
## Description
JSON file validation in the file Editor.


ehynd:
image


@KirthikaKumar-K can we change this error message to remove the second sentence? Just say like "This file does not contain valid JSON" since the user can still upload the file regardless of it being buggy

KirthikaKumar-K:
> his file does not contain valid JSON

Hi @ehynd 
I have fixed the error message according to your suggestion.
https://github.com/user-attachments/assets/a191642d-d70f-4a87-ad75-213bcfad1dc1

neelneelneel:
Not necessary in the new component. User should upload proper json if they want to use it

PR #2460, opened 2025-12-19
Fortify issue fixes

soumya-pratik:
## Description
Added fixes for Fortify threats


packages/playground/src/pages/workspace-detail-page.tsx
```diff
@@ -128,7 +139,13 @@ export const WorkspaceDetailPage = observer(() => {
 						<DropdownMenuContent align="end">
 							<DropdownMenuGroup>
 								<DropdownMenuItem asChild>
-									<Link to={`/workspace/${workspaceId}/edit`}>
+									<Link
+										to={
+											safeWorkspaceId
+												? `/workspace/${safeWorkspaceId}/edit`
```
neelneelneel: Use the ID returned from getWorkspace. There is no need to do validation on the frontend.

packages/client/src/utility/general.ts
```diff
@@ -249,3 +249,48 @@ export const extractInitials = (str: string): string => {
...
+	// Handle null/undefined inputs
+	if (!url || typeof url !== 'string') {
+		return 'about:blank';
+	}
+
+	// Trim whitespace
+	const trimmedUrl = url.trim();
+	
+	if (trimmedUrl.length === 0) {
+		return 'about:blank';
```
neelneelneel, line 269: Why are you returning about:blank?

neelneelneel (changes requested):

packages/client/src/utility/general.ts
soumya-pratik: @neelneelneel the purpose of this function is to check for valid urls and for any invalid/malicious url I am returning a blank navigation. If you have any better suggestion to handle this scenario please let me know , I will update it.

packages/playground/src/pages/workspace-detail-page.tsx
NayanKumar809: Reverted the validation here.

PR #2532, opened 2026-01-15
Issue with layers search bar and auto scroll

### LINKED ISSUE #566: Drag and drop app: Issues with layers search bar

1. Open the application
2. Create Drag and drop app
3. Drag and drop some blocks on page(container, Button, Text, etc)
4. Clicks on Layers option
5. Observe the search bar is present
6. Duplicate the blocks multiple times
**Issue 1:**
7. Search bar is fully outside the viewable area.
8. Reduce the size of window then search bar displayed
**Issue 2:** 
9.  Search the layers name for which scroll needs.
10. Searched layer is highlighted but needs to scroll

### Expected Result

1. The search bar should always be visible above the list of layers, regardless of the window size.
2. After searching, the list should automatically scroll to the highlighted matching layer(s) so the result is clearly visible without manual scrolling.

### Actual Results

1. The search bar is hidden when multiple layers are added and is only displayed upon reducing the window size.
2. Only the matching layer(s) are highlighted, but the user must manually scroll through the list to locate the result, making the search functionality inefficient.

### Additional Context

https://github.com/user-attachments/assets/42b274e7-2f57-4bf8-9804-8ffa06dfb2f2

### Requirements or Tasks

_No response_

- [x] Yes

- [x] Yes

Vishal24-Kanini:
## Description
Resolved the issues related to search bar view and auto scroll and highlight the searched item in layers panel

## Changes Made

image

image

## How to Test

2. Expected outcomes


Paulson-Robert (approved):
Approved. Should start migrating it into shadcn

PR #2535, opened 2026-01-19
feat: add model id to room options in playground

rithvik-doshi:
Added modelId to playground room options

To test:
- Start a new chat with playground and ensure modelId is added to the room options.
- Then during a chat, switch to another model and ensure that, when update room options is called the modelId also switches in the room options.
- Before regenerating a message, switch the model. The modelId in the room options should also update accordingly.


tevanburen (commented):
We should probably just add modelId to room options, then use that modelId to set model when loading a room

tlokeshrao:
Works as per PR description

tlokeshrao (approved):

PR #2538, opened 2026-01-20
Resolve 572: 'Data grid Frame header not refreshing with new frame'

### LINKED ISSUE #572: BUG: Data grid Frame header not refreshing with new frame

1. Create a Drag & Drop application. 
2. Run a query to create two frames with two different data sets.
3. Drag a Data Grid block onto the page
4. Select a frame from the dropdown and manually refresh the headers
5. Select the 2nd frame
6. Try to refresh the headers

### Expected Result

The expected results are that the headers should refresh to the new frame in the drop down.

### Actual Results

You get the following error: "Converting circular structure to JSON --> starting at object with constructor 'Object' --- property 'ADOUserId' closes the circle."

### Additional Context

_No response_

### Requirements or Tasks

- [ ]  Remove previous headers when selecting a new frame
- [ ] Have the frames automatically refresh when selecting a frame. 

- [x] Yes

- [x] Yes

JoshRome26:
## Description
While editing the data grid block and you change the frame in the Data tab, it will now automatically clear the previous headers and reload the grid to the new frames headers.
Updated the grid block column settings to automatically sync frame headers when changing frames while editing the data grid block.

## Changes Made
Updated the grid block column settings:
- When creating the new column map, change to assign to the column for each key instead of the object itself. This was causing errors when trying to reload manually if previous headers were present. 
- Add useEffect() to call the syncHeaders() function when frame changes occur.
- Add state change to clear out columns when changing the frame. 
- Remove key from GridBlockSettingsItem to resolve error.

## How to Test

1. Create a Drag and Drop application and go into Edit mode.
2. Add notebook with two cells that contain different frame data.
3. Create a Grid Block and assign the first newly created frame. 
4. Headers should automatically sync on the block with a notification appearing on the top right. 
5. Change the frame to the 2nd.
6. Headers should clear, with new headers from the frame appearing and the same notification.

### DIFF
--- a/packages/client/src/components/blocks-workspace/blocks/settings/custom/grid-two/GridBlockColumnSettings.tsx
+++ b/packages/client/src/components/blocks-workspace/blocks/settings/custom/grid-two/GridBlockColumnSettings.tsx
@@ -8,6 +8,7 @@ import {
 import { CSS } from "@dnd-kit/utilities";
 import { Sync } from "@mui/icons-material";
 import { observer } from "mobx-react-lite";
+import { useEffect } from "react";
 import {
 	type GridBlockColumn,
 	type GridBlockDef,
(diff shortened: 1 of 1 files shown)


JoshRome26:
New UI framework adoption fixed this issue.

PR #2566, opened 2026-01-23
fix: adding unzip as secondary option on zip files

ppatel9703:
## Description
Adding in secondary option for zip files to allow for unzip

## Changes Made
adding in new secondary option in the app-file-expoloer
## How to Test

1. create new app 
2. unpload zip 
3. hit secondary menu to unzip 
image
image
image


ppatel9703:
Duplicate PR for same issue

PR #2606, opened 2026-01-30
Update Vega functions and remix router package

JaganSomannaKanini:
## Description
In this PR updating of the vega function and remix-run/router package is done.

## Changes Made
Added the following in the root package.json
"pnpm": {
      "overrides": {
          "@remix-run/router": "1.23.2",
          "vega-functions": "6.1.1",
          "vega-parser>vega-functions": "6.1.1",
          "vega-view>vega-functions": "6.1.1",
          "vega>vega-functions": "6.1.1"
      }
  }

## Notes
@johbaxter 
- The @remix-run/router and vega-functions, which are transitive dependencies (not directly used in our source code). Installing these packages directly was not appropriate because it would only add unused top-level versions, while the vulnerable versions inside react-router and vega would still be used at runtime.
- Upgrading the parent libraries (react-router, react-router-dom, vega, react-vega) did not fix the issue because their latest compatible releases still depend on the vulnerable sub dependency versions.
- I resolved this by using pnpm overrides, which forces all packages in the dependency tree to use the secure versions.

The following overrides were applied:

@remix-run/router → 1.23.2
vega-functions → 6.1.1

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,5 +33,14 @@
   "engines": {
     "pnpm": "~10.13.0",
     "node": ">=24.4.0 <25"
+  },
+  "pnpm": {
+      "overrides": {
+          "@remix-run/router": "1.23.2",
+          "vega-functions": "6.1.1",
+          "vega-parser>vega-functions": "6.1.1",
+          "vega-view>vega-functions": "6.1.1",
+          "vega>vega-functions": "6.1.1"
+      }
   }
 }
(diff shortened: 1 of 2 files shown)


neelneelneel (changes requested):
What are the implications of updating the base packages? We shouldn't override.

JaganSomannaKanini:
> What are the implications of updating the base packages? We shouldn't override.

@neelneelneel I have updated the base packages, it's been pushed to a new PR https://github.com/SEMOSS/semoss-ui/pull/2681. There were some issues on updating this PR so created a new one so closing this.

PR #2651, opened 2026-02-05
fix: pass workspace id to create playground room

rithvik-doshi:
To ensure we properly link rooms with workspaces.

BE PR: https://github.com/SEMOSS/Semoss/pull/2039


tevanburen (approved):
Works

PR #2674, opened 2026-02-10
fix for overflow text for long prompts

ekim1707:
## Description
Made html and style changes to address text overflow when inserting long prompts into room input

## Changes Made
made parent wrapper and other style changes


neelneelneel (changes requested):
This effects the click propagation. Instead can you

1. Add a background to the settings / microphone, so the background text doesn't go through
2. Add a listener to scroll to the bottom of click

ekim1707:
Major changes:
1. Editor card parent (wrapper) is relative and group, allowing scrim and icon rows (absolute) to anchor to it instead of to the entire page
2. Scroll behavior was fixed so only one container has scroll (min-h-0 flex-1 overflow-y-auto), preventing children inside from performing scrolling behavior in odd ways
3. Padding added to text caret doesnt go underneath menu row and added solid scrim background to prevent text from bleeding out from underneath row.

neelneelneel:
Hey this is still not correct. A few issues:
- [ ] Click passed in from new-room-page on on the input was removed, this is necessary for other themes
- [ ] Clicks on the bottom bar do not propagate to the underling input
- [ ] CSS is off
   - [ ] Hover / Drag css is not applied the same way 
   - [ ] Corners are getting cut off
- [ ] Resolve biome formatting/linting issues

neelneelneel (approved):

PR #2721, opened 2026-02-18
fixed error message

ekim1707:
## Description
Changed error message.

## Changes Made
Message wording

## How to Test
1. Upload an image larger than 5MB
2. Enter a prompt ("describe the image") and press enter

### DIFF
--- a/packages/playground/src/components/room/room-content.tsx
+++ b/packages/playground/src/components/room/room-content.tsx
@@ -263,6 +263,28 @@ export const RoomContent: React.FC<RoomContentProps> = observer(({ room }) => {
 		};
 	}, [contentEle]);
 
+	/**
+     * Error handling
+     */
+    const getErrorMessage = (error: any): string => {
+		const errorString = error.toString().toLowerCase();
+		if (errorString.includes("5 mb")) {
+			return "One of your files exceeds the 5 MB limit. Please upload smaller files and try again.";
+		}
+
+		// Needs refactor bc error is a string, not an object:
+		
+		// Check for error code in different structures 
+		// const code = error?.error?.code || error?.status || error?.statusCode;
+		
+		// if (code === 429) {
+		// 	return "This model is experiencing high load right now. Please try a different model.";
+		// }
+
+		// Fallback to original message or default
+		return "Unable to process request. Please check your connection, copy your message, and refresh.";
+    };
+
 	return (
 		<div className="flex h-full w-full flex-col bg-secondary-background transition-all duration-200 ease-in-out">
 			<div className="relative w-full flex-1 overflow-hidden">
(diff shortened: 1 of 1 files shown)


ekim1707:
pulling this pr back bc addressed by another ticket.

ekim1707:
@ckelly17 hoping to include this in the next merge to address ticket [693 ](https://sde.fda.gov/jira/browse/ELSA-693)

packages/playground/src/components/room/room-content.tsx
```diff
@@ -327,8 +349,7 @@ export const RoomContent: React.FC<RoomContentProps> = observer(({ room }) => {
 									<TriangleAlertIcon className="h-6 w-6" />
 								</div>
 								<span>
-									{room.error.message ||
-										t("content.errorDefault")}
+									{getErrorMessage(room.error)}
```
neelneelneel, line 352: This breaks i18n

packages/playground/src/components/room/room-content.tsx
```diff
@@ -263,6 +263,28 @@ export const RoomContent: React.FC<RoomContentProps> = observer(({ room }) => {
 		};
 	}, [contentEle]);
 
+	/**
+     * Error handling
+     */
+    const getErrorMessage = (error: any): string => {
+		const errorString = error.toString().toLowerCase();
+		if (errorString.includes("5 mb")) {
```
neelneelneel, line 271: This is not valid, there are many edge cases and different types of error. How will you process token limit errors? guardrail errors?

neelneelneel (changes requested):
Please think through and validate your work before submitting a PR

packages/playground/src/components/room/room-content.tsx
ekim1707: Hey Neel, i was asked to resolve the use case of image size > 5 MB (as the Jira ticket shows) as somewhat of an urgent fix, so I thought I would begin with that and then we could build the error handling from there. Regardless, I apologize for that and I'll try to make the changes you asked for asap

ckelly17: Need to make error message more specific if it’s just going to handle this particular case.

There are a number of other error types that need to be handle but outside of this case. What can we do to handle this specific case? Might even want a preflight check on image size to prevent getting this error in the first place

This graceful handling of back-end errors will be a theme: prompt too long, network errors, resource exhausted

ekim1707: Think it would be best to group up and decide how we do this. One thing i wasn't sure of was that all the errors are coming back as strings, which makes it a little difficult to navigate.

ekim1707:
no longer needed

PR #2726, opened 2026-02-19
potential fix- needs testing with Windows cpu

ekim1707:
## Description
Most significant change is that If both plain text and image exist in non-html or html clipboard data, only text will be pasted.

## Changes Made
Instead of only looking at the uploaded files, the clipboard data is examined for plain image and text as well as html image and text. Certain times, such as with Excel text, html images were included in the clipboard data.

## How to Test (need Windows computer)
1. Try to copy/paste any image into the textarea. The image box should appear underneath.
2. Try to copy/paste any text from Excel, OneNote, etc. Only the text should appear in the textarea without any image box underneath.

## Notes
Should look something like this:
image (1)
 - Excel test pasted successfully without any associated image
 - Real Madrid image pasted successfully

### DIFF
--- a/packages/playground/src/components/room/room-input.tsx
+++ b/packages/playground/src/components/room/room-input.tsx
@@ -355,19 +355,51 @@ export const RoomInput: React.FC<RoomInputProps> = observer(
 											setIsDragging(false);
 										}}
 										onPaste={(e) => {
-											// set the new files
-											const updated = Array.from(
-												e.clipboardData.files,
+											const cd = e.clipboardData;
+
+											const imageFiles = Array.from(cd.files || []).filter((f) =>
+												f.type?.startsWith("image/")
 											);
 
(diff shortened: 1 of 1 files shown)


neelneelneel:
Why not pass the clipdata directly to the files? When we upload, we process the files to split out images.

ekim1707:
> Why not pass the clipdata directly to the files? When we upload, we process the files to split out images.

@kzsb03 could you look over this with me? thanks

kzsb03:
Sure feel free to ping

ckelly17:
@kzsb03 @ekim1707 what is the status on this?

ekim1707:
> @kzsb03 @ekim1707 what is the status on this?

need Kyle to take a look at this before proceeding

ekim1707:
no longer needed

PR #2731, opened 2026-02-20
Page 2 is visible on Page 1 after navigation in drag-and-drop app bug fix

### LINKED ISSUE #2584: Page 2 is visible on Page 1 after navigation in drag-and-drop app

1. Open the application (drag-and-drop app).
2. Navigate to the Layers tab.
3. Add a new page (Page 2) in the layer.
4. Move back (navigate) to Page 1.
5. Observe the pages displayed in the page screen (workspace).

### Expected Result

After navigating to Page 1, only Page 1 and its relevant content should be displayed.
Page 2 should not be visible or accessible from the view or workspace of Page 1.

### Actual Results

After navigating to Page 1, Page 2 is still visible/present in the page screen.
This creates confusion and breaks the isolation expected between pages.

### Additional Context

Image

### Requirements or Tasks

_No response_

- [x] Yes

- [x] Yes

KirthikaKumar-K:
## How to Test

2. Expected outcomes


neelneelneel (changes requested):
Workspace is used in multpile places and designer may not always be in the context tree. I recommend moving this to a lower level, possibly `BlocksWorkspace`

KirthikaKumar-K:
> Workspace is used in multpile places and designer may not always be in the context tree. I recommend moving this to a lower level, possibly `BlocksWorkspace`



> Workspace is used in multpile places and designer may not always be in the context tree. I recommend moving this to a lower level, possibly `BlocksWorkspace`

We will look into this Neel.

Paulson-Robert (approved):
As Neel suggested, the designer check prospect is moved up a layer.

Paulson-Robert (approved):

johbaxter (approved):

PR #2737, opened 2026-02-20
fix prevents removing of dropdown menu items after chat

ekim1707:
## How to Test
There are 2 flows that need testing:
1. Go to new room page and select an agent, knowledge base, and toolbox.
2. Enter a prompt
3. In the following room page, check to see if the agent name, knowledge base, and toolbox are still selected and showing on the dropdown menu.

Also:
1. Start from agent page, click on an agent, and click on "new chat"
2. Follow the same steps as above and make sure everything is still showing upon entering the created room page as well.


neelneelneel (changes requested):
@kunal0137 - is this a valid and approved requirement? If so, it should be done on the backend.

ekim1707:
no longer needed

PR #2756, opened 2026-02-24
feat: reduce default size of the side panel and configure min width

rithvik-doshi:
## Description
Configure the min and max panel width and set it to 50% initially.

## How to Test
1. Open and play with the side panel


kzsb03 (changes requested):
Hesitant on max size

rithvik-doshi:
@kzsb03 Could remove max size - but does the default + min look fine?

bbelledd (approved):

kzsb03 (approved):

PR #2764, opened 2026-02-25
680 add tools page

Kaveh-Vakili:
## Description
Add a new Tools page to the playground sidebar that displays MCP tools grouped by their parent app/project.

## Changes Made
- Created `tools-page.tsx` in `src/pages/` with search bar, filter buttons, and card grid layout
- Created `ToolCard.tsx` in `src/components/` that fetches MCP tools per project using `GetMCPTools` and renders individual tool cards with tool name, description, parent app name, and visibility badge
- Added `/tools` route to `router.tsx`
- Added Tools link with WrenchIcon to `global-nav.tsx` sidebar
- Exported ToolCard from `src/components/common/index.ts`
- Data fetched using `MyEngineProject` (for apps) and `GetMCPTools` (for tools per app)

## How to Test
1. Run `pnpm run dev` and navigate to `localhost:5174/#/tools`
2. Verify "Tools" appears in the left sidebar with a wrench icon
3. Verify the Tools page loads with title, subtitle, search bar, and filter buttons
4. Verify tool cards display with tool name, description, parent app name, and PUBLIC/PRIVATE badge

## Notes
- Tools data depends on MCP projects existing in the system
- Search uses `filterWord` parameter on `MyEngineProject` pixel query
- Card visibility badge is based on `project_global` field from the API

Screenshot 2026-02-26 101059


bbelledd (changes requested):
Image

**1.** This header is abnormally large to our other sizes.  I would suggest scaling down to  `font-bold text-4xl` instead of 5-xl

**2.** Our UI kit has its own tab components. We don't use buttons as navigation. Please refer to [this documentation](https://ui.shadcn.com/docs/components/radix/tabs) on how to use the tab component appropriately and replace these changes.

**3.** This element is a _badge_, but does not utilize our shadcn badge component. Please refer to [this documentation](https://ui.shadcn.com/docs/components/radix/badge) on how to use the badge component appropriately and replace these changes.

**4. & 5.** Padding between the div is uneven and extremely large. If many tools were to populate this screen, the current width of these elements are limiting view of compacting more tools into the view height. I would recommend referencing the cards used in the Agents tab and mimic the similar 24px padding used all around the div. 

**6.** Within your `<div className="relative flex h-full w-full flex-col gap-4 overflow-hidden pl-2"></div>` element, the overarching padding is not consistent with our standard 16px padding, aka `p-4`. Currently you only implemented padding `pl-2`, please fix this to a right and left padding of `p-4` . See blue annotation in image for reference of how the padding should look. 

**7.** The background color should not inherit the original gradient background we have. Please reference the Agents tab and match the flat background color variable there.

Kaveh-Vakili:
Pushed all changes in relation to the comments listed above

PR #2770, opened 2026-02-26
Add folder zip download in room explorer

ckelly17:
## Description
Adds ability to download a zip file of a directory in room explorer. This saves the user from needing to download files individually if there are many. This is most helpful if a process (e.g., an MCP creating documents) has created or placed a number of files into the room.

## Changes Made
Removes directory exclusion from `packages/playground/src/components/room/room-file-explorer.tsx` and exposes download options for directories

## How to Test
1. Upload files/directories into the room via the room explorer
2. Hover over three dots on right to see download option. 
3. Click download

image

**Result**
image


SunTheDev:
Reviewed and tested the functionality which works as intended. Able to see the download button and can download a zip folder.

PR #2772, opened 2026-02-26
Render HTML preview in room explorer

ckelly17:
## Description
Adds HTML file preview support to the shared file explorer. When a user clicks an .html or .htm file, it now renders in an inline preview instead of opening directly in the Monaco code editor. The main benefit of this feature is to allow the model (via an MCP) to create html-based graphics and UIs and make them immediately accessible to the user, rather than requiring them to download the html and open in browser.

image

## Changes Made
- `libs/shared/src/components/file/file-html-viewer.tsx (new file`) — New FileHtmlViewer component that fetches the file's text content and renders it in an <iframe srcDoc>. Includes a toolbar with Preview/Source toggle buttons and an "Open in new tab" button that creates a temporary Blob URL.
- `libs/shared/src/components/file/file-editor.tsx` — Added isHtml check for .html/.htm extensions to route those files to FileHtmlViewer instead of FileCodeEditor.
- `libs/shared/src/components/file/index.ts` — Exports FileHtmlViewer.

## How to Test
- Open the file explorer in playground room
- Upload or navigate to an .html file and click it
- Verify the file renders as HTML in the Preview tab (styles, layout, and JavaScript should all work)
- Click Source to confirm the raw HTML is shown in the Monaco editor with full edit/save/download functionality
- Click Preview to switch back — content should already be loaded (no re-fetch)
- Click the external link icon to open the rendered HTML in a new browser tab; confirm it renders correctly and the tab opens without a download prompt


libs/shared/src/components/file/file-html-viewer.tsx
```diff
@@ -0,0 +1,120 @@
...
+								<Muted className="text-destructive">
+									{getFile.error?.message ||
+										"Failed to load file"}
+								</Muted>
+							</div>
+						)}
+						{getFile.status === "SUCCESS" && (
+							<iframe
+								className="h-full w-full border-0"
+								srcDoc={getFile.data}
```
neelneelneel: Vulnerable to XSS. See https://developer.mozilla.org/en-US/docs/Web/API/HTMLIFrameElement/srcdoc

Need to look into a sandboxed way to run code. Recommend looking into packages that can render react in the browser. as well

neelneelneel (changes requested):

neelneelneel:
Still vulnerable, will not pass a security scan

PR #2779, opened 2026-03-01
Handle links to room explorer artifacts (Playground)

ckelly17:
## Description
Creates a pathway and handling mechanism for playground model to create clickable links to artifacts in room explorer. This is most useful for citations and linking to generated outputs.

## Changes Made
- Override URL sanitization in `libs/ui/src/next/markdown.tsx` to allow passing a custom prefix ("room://")
- Add custom link handling in `packages/playground/src/components/message/response-message-text.tsx` and `packages/playground/src/components/message/response-message.tsx` to allow links starting with `room://` to open side panel instead of navigating out

## How to Test
1. Upload a file to room explorer
2. Instruct model on how to create a link to that location. Ideally, this would be in a system prompt or some generalized instructions, but you can reproduce with a manual prompt (e.g., `Provide citations in your response by prefixing "room://" to the hyperlink you create to the file path`)
3. Click on link in chat. It should open the file explorer to the cited artifact.

**Prompt**
image

**Result**
image

## Notes
This feature relies on specific prompting instructions to the LLM. It is not a generic solution. If links are rendered without the `room://` prefix, current click behavior should not change

image


packages/playground/src/components/message/response-message.tsx
```diff
@@ -93,6 +117,7 @@ export const ResponseMessage: React.FC<ResponseMessageProps> = observer(
 										message={message}
 										part={p}
 										isLast={isLast}
+										onRoomLink={handleRoomLink}
```
neelneelneel: No need to have an additional prop, you can trigger the method directly in response-message-text.

neelneelneel (changes requested):

packages/playground/src/components/message/response-message.tsx
ckelly17: Additional prop removed

kzsb03 (approved):

kzsb03 (approved):

PR #2798, opened 2026-03-03
fix: infinite loop on error

tevanburen:
## Description
Some pages show infinite loops when reactors error

## Changes Made
Calling notification.add was changing the value of notification.add, which can cause an infinite loop in a useEffect. Wrap notification.add in a callback

## How to Test
Many ways, but here's one:
1. Get on SemossWeb dev
2. In the GetProjectDependencies reactor, have it always throw an exception (for example, add `int tmp = 5 / 0;`)
3. Restart your server
4. On the App info page (where you would normally view dependencies), notice that GetProjectDependencies is called infinitely
5. Get on this branch
6. Rebuild libs/ui and packages/client
7.  On the App info page, verify that the error is shown but without bricking the page

### DIFF
--- a/libs/ui/src/components/Notification/Notification.tsx
+++ b/libs/ui/src/components/Notification/Notification.tsx
@@ -1,6 +1,12 @@
 import { Snackbar, type SxProps } from "@mui/material";
 import type React from "react";
-import { isValidElement, type ReactNode, useEffect, useState } from "react";
+import {
+	isValidElement,
+	type ReactNode,
+	useCallback,
+	useEffect,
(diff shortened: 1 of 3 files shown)


spatel776 (approved):
Have just tried the test and it works. Previously, with injecting a RunTime error in `GetProjectDependencies`, it would brick the page and prevent me from entering another page. With pulling and building this branch, I was able to have an error show, but did not prevent me from entering other tabs on the App Info.

kzsb03 (approved):

PR #2804, opened 2026-03-04
Optional theme to disable links to platform from Playground

ckelly17:
## Description
This PR provides a means to hide the various links to the base platform for engines, agents, MCPs, etc. It implements conditional logic set in the theme database to hide these links if desired. If no value is set in theme, links will be shown.

## Changes Made
- New theme value set in `libs/shared/src/types.ts` and stored in `packages/playground/src/stores/root/root.store.ts`
- Conditional logic set in five other places

## How to Test
These links appear in four visible spots:

1. Agent detail page (links to vectors and MCPs)
2. New chat settings page when selecting knowledge and MCPs
3. New chat agent dropdown menu
4. Selecting "new chat" from agent page renders badge with link on new room page

In order to disable these, set `{"showPlatformLinks": false}` in theme. Confirm that none of these locations display the links. To test the reverse, either set to `true` or have no value for `showPlatformLinks` in theme. All links should appear as they currently do.

**Example Shown (Agent Detail Page)**
image

**Example Hidden**
image

## Notes
The theme setting `showPlatformLinks: false` hides external links to the SEMOSS platform. It's read directly in each of the 5 components that render such a link, via:
`root.theme.showPlatformLinks !== false`
The check is duplicated in each file. It works, but if someone adds a new platform link in the future they have to know to add the same check manually.

_Possible future approach (not yet implemented, need consensus):_
Create a <PlatformLink href="..."> wrapper component in `packages/playground/src/components/common/`. It would internally do the `showPlatformLinks` check and either render an `<a target="_blank">` or return null. Any new platform link would just use this component instead of a bare `<a>`, and the gating behavior would come automatically be applied. The convention would be: if the link goes to the SEMOSS platform, use `<PlatformLink>`.

The five existing locations would need to be refactored to use it — that work hasn't been done yet.


kzsb03:
@ckelly17 Should this include ability to customize routes if enabled e.g. linking to playground knowledge page

ckelly17:
@kzsb03 I like that idea. Let's talk about best way to implement

kzsb03:
@ckelly17 We should default the value to true in the root store, currently not set

ckelly17:
@kzsb03 updated and confirmed that works as scoped when `showPlatformLinks` is set to `true` or `false` in theme, or if omitted. Only `false` explicitly hides the link
image

neelneelneel:
I don't think this is valid yet. Base platform has functionality that is missing from the playground and is important in understanding the app/engine.

neelneelneel:
I don't think this is valid yet. Base platform has functionality that is missing from the playground and is important in understanding the app/engine.

PR #2836, opened 2026-03-10
Remove filter on MCP tag if knowledge MCP disabled (playground)

ckelly17:
## Description
Removes filter on `MCP` tag when knowledge MCPs are disabled via theme. This occurs on agent creation page and knowledge selection on main chat input panel.

If knowledge MCP is enabled (as is default), there is no change.

## Changes Made
Updates to `packages/playground/src/components/mcp/mcp-selector.tsx` and `packages/playground/src/components/room/room-input-menu-knowledge.tsx`

## How to Test
In your theme DB set `"enableKnowledgeMCP" : false` and check:

1. Network calls show no filter on MCP tag on MyEngines query for vectors
2. Any non-MCP vectors are returned

For regression testing, set `"enableKnowledgeMCP" : true` or omit the value from theme. Both places should return only vectors tagged with `MCP`. 

## Notes
The `MCP` tag filter is not applied on the main /knowledge page, and so no update is made there.


packages/playground/src/components/mcp/mcp-selector.tsx
```diff
@@ -81,7 +84,7 @@ export const MCPSelector: React.FC<MCPSelectorProps> = ({
 	 */
 	const getMCP = useIteratorPixel<(Engine | App)[], MCP>(
 		(limit, offset) =>
-			`MyEngineProject (metaKeys = ["tag", "description"], metaFilters=[{"tag":["MCP"]}], type=${type === "TOOLBOX" ? `["PROJECT", "STORAGE", "DATABASE", "FUNCTION"]` : `["VECTOR"]`}, ${debouncedSearch ? `filterWord=${JSON.stringify(debouncedSearch)}, ` : ""}limit=[${limit}], offset=[${offset}])`,
+			`MyEngineProject (metaKeys = ["tag", "description"], ${useMCPFilter ? `metaFilters=[{"tag":["MCP"]}], ` : ""}type=${type === "TOOLBOX" ? `["PROJECT", "STORAGE", "DATABASE", "FUNCTION"]` : `["VECTOR"]`}, ${debouncedSearch ? `filterWord=${JSON.stringify(debouncedSearch)}, ` : ""}limit=[${limit}], offset=[${offset}])`,
```
kzsb03, line 87: We still need to have MCP engines come back for non-knowledge

kzsb03 (changes requested):

kzsb03 (approved):

packages/playground/src/components/mcp/mcp-selector.tsx
ckelly17: Confirmed they do

PR #2853, opened 2026-03-13
feat(playground): add key for alternate landing page title

travon-speller:
## Description

Adds new theme key for alternate landing page title

## How to Test

1. update admin theme with new keys
2. go to playground and add altLandingKey into the url
3. see alternate title


tevanburen:
Out of curiosity, what's the use case?

PR #2892, opened 2026-03-20
Sidebar Expanded by Default- 881

SunTheDev:
## Description
Updated initial state of sidebar to true so whenever user comes to the playground app for the first time, it will be expanded.

## Changes Made
isSidebarOpen state is now based on themedb theme.sidebar. expandedByDefault, requires expandedByDefault = true

## How to Test

1. Clear your cache
2. Open the app locally, make sure you have expandedByDefault set to true in themedb, and confirm your app has sidebar expanded.


packages/playground/src/pages/main-layout.tsx
```diff
@@ -30,7 +30,7 @@ export const MainLayout = observer(() => {
 	const [navbarActions, setNavbarActions] = useState<ReactNode | null>(null);
 
 	const [isSidebarOpen, setIsSidebarOpen] = useCacheState(
-		false,
+		theme?.sidebar?.expandedByDefault ?? false,
```
kzsb03: this should be defined on root store as a default in the type for theme and have a default value so users know its an option

Kaveh-Vakili:
Reviewed code with Sunny and functionality works as described. The changes are well implemented.

kzsb03 (approved):

PR #2902, opened 2026-03-24
UI chat file icons

Kaveh-Vakili:
## Description
Refactored the file upload display in the chat room input to show image thumbnails for uploaded images, distinct monochrome icons with extension labels for non-image files, and a horizontal carousel with scroll arrows when many files are attached.

## Changes Made
Image previews: PNG, JPG, JPEG, GIF, WEBP, and SVG files now display an actual image thumbnail using URL.createObjectURL, with proper cleanup via useEffect to prevent memory leaks
Unique file type icons: Each file type category has its own distinct lucide icon (e.g., FileSpreadsheetIcon for spreadsheets, FileTerminalIcon for shell scripts, FileBadgeIcon for PDFs, FileArchiveIcon for archives, etc.)
Monochrome styling: All file icons use text-muted-foreground with a reduced strokeWidth of 1.25 for a lighter, cleaner appearance
Extension labels: Non-image files display an uppercase extension label (e.g., "PDF", "CSV", "PPTX") beneath the icon so the file type is immediately visible without hovering
Carousel navigation: When uploaded files overflow the container, left/right chevron buttons appear to scroll through the file list. The scrollbar is hidden and scrolling is smooth
Fixed file removal: Moved the X (remove) button outside the TooltipTrigger so click events are no longer swallowed by the tooltip, fixing the issue where files could not be removed

## How to Test
Navigate to a chat room input
Upload one or more image files (PNG, JPG, etc.) — verify they show as image thumbnails
Upload non-image files (PDF, CSV, PPTX, TXT, PY, ZIP, etc.) — verify each shows a distinct grey icon with the file extension label beneath it
Upload enough files to overflow the container width — verify left/right scroll arrows appear and work
Hover over a file and click the X button — verify the file is removed from the list
Submit a prompt with files attached — verify files are cleared after successful submission

## Notes
The FileImageIcon import was removed as image files now show actual previews instead of a generic icon
Object URLs are managed with useMemo and revoked on cleanup to avoid memory leaks
Icon stroke width is set to 1.25 (down from the default 2) for a thinner, less heavy 

image


ckelly17:

SunTheDev:
File icons are shown as described, a few items might need attention (Tested locally):

1) Attaching both images and documents, only images show up

2) Attached 1 pdf and asked a question, AskPlayground doesn't contain reference to doc and response is that it can't see a doc attached.

3) Duplicate images can be attached:
1. Upload image 1
2. Upload image 2
3. Upload image 1 again and see its attached

ckelly17:
@SunTheDev please confirm that:

1. What do you mean by "show up"? are previewed?
2. Check that your theme DB isn't filtering out PDFs from AskPlayground call
3. Is that the behavior on `dev` branch, or specific to this one?

kzsb03 (approved):

PR #2903, opened 2026-03-24
resolved copy from microsoft apps shows as image bug

Kaveh-Vakili:
## Description
Fixes an issue where copying text from Microsoft applications (Word, PowerPoint, Excel, OneNote, etc.) and pasting into the chat input would incorrectly attach an image file instead of pasting the text content.

## Changes Made
Updated the onPaste handler in room-input.tsx to detect when the clipboard contains both text and image data
When text content (text/plain or text/html) is present alongside image files, the image files are filtered out so the text pastes normally
Pure image pastes (e.g., screenshots) still attach as files as expected

## How to Test
Copy text from a Microsoft app (Word, Outlook, Teams)
Paste into the chat input
Expected: The text is pasted as text, no image file is attached
Take a screenshot or copy an image directly
Paste into the chat input
Expected: The image is attached as a file

## Notes
Microsoft apps include an image representation of copied content in the clipboard alongside the text, which was being picked up by the file handler


ckelly17:

PR #2913, opened 2026-03-26
fix: override faltted version to 3.4.2

soumya-pratik:
## Description
As flatted@3.3.3 is a transitive dependency — it's pulled in by:

@vitest/ui (via @vitest/ui@4.0.14)
flat-cache (via flat-cache@4.0.1)

This particular version has been flagged for having critical vulnerability. 
So as a fix i have added an override to use 3.4.2 version instead.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,5 +33,10 @@
   "engines": {
     "pnpm": "~10.13.0",
     "node": ">=24.4.0 <25"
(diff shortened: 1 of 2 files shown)


johbaxter (approved):
Pulled builds locally across packages and libs

PR #2918, opened 2026-03-26
fix: add name behind weclome in new room

yifan-z0698:
## Description
add user name behind the welcome in the new room page

## Changes Made
new-room-page.tsx

## How to Test
the playground new room page, welcome text

## Notes
NA


ckelly17:
@yifan-z0698 could you add a screenshot? Thank you!

yifan-z0698:
> @yifan-z0698 could you add a screenshot? Thank you!

Here is the screenshot.
image

ckelly17:
Thank you!

tevanburen (changes requested):
Thanks for the PR! This looks good, and I think we can enhance it using the semoss-ui/libs/i18n folder.

I'm concerned that when we add multi-language support, in some languages the user's name should go _before_ the "welcome" word. Like "Thomas, Welcome" instead of "Welcome, Thomas".

Our i18n lib lets us use placeholder variables, like `"Welcome, {{name}}"`. See `memoryUsedValue` in room.json for an example. Rather than doing this string concatenation in new-room-page, could you add name as a placeholder variable in room.json, and then pass it new-room-page?

If you do the english one, I can take care of the other languages - just let me know. Thanks again

yifan-z0698:
Thank you, Thomas for the comment.

Base on your suggestion, I've updated ```room.json``` for all languages, using ```{{name}}``` as a placeholder variable.

For most languages other than English, the order is Welcome, Name:
- 🇸🇦 Arabic: "مرحباً، {{name}}"
- 🇪🇸 Spanish: "Bienvenido, {{name}}"
- 🇫🇷 French: "Bienvenue, {{name}}"

For languages where the name naturally comes first (Name, Welcome):
- 🇮🇳 Hindi: "{{name}}, स्वागत है"
- 🇯🇵 Japanese: "{{name}}さん、ようこそ"

Correct me if I got any of them wrong and let me know if any of the translations need adjusting.

tevanburen (approved):
Awesome, thank you

PR #2921, opened 2026-03-27
fix: sast issues

soumya-pratik:
## Description
Fix -
1. HTML5: Overly Permissive Message Posting Policy


johbaxter (approved):
Looks good to me, just limiting where we post tool calls (security vulnerability).  Only allow post messages to same domain.  Will merge soon after one more approval.

PR #2942, opened 2026-03-30
scroll to bottom of the chat

Kaveh-Vakili:
## Description
When selecting a chat from history, the conversation view would start at the top of the message thread, requiring the user to manually scroll down to the most recent messages. This fix automatically scrolls to the bottom of the conversation when a room is loaded from history.

## Changes Made
packages/playground/src/components/room/room-content.tsx: Added a useEffect that triggers when the active room changes. It resets the scroll lock (isScrollLocked) and programmatically scrolls the message container to the bottom after a short delay to allow content to render.

## How to Test
1. Open the playground with an existing chat history
2. From the chat list/sidebar, select a previous conversation that has multiple messages
3. Expected: The conversation view should automatically scroll to the most recent message at the bottom
4. Previously: The conversation would open at the top of the thread, requiring manual scrolling
5. Verify that sending a new message in an active room still auto-scrolls correctly
6. Verify that manually scrolling up in an active conversation (scroll lock behavior) still works as expected


ckelly17:

kzsb03 (approved):

PR #2945, opened 2026-03-30
CSV Download Button for Run Query Results

### LINKED ISSUE #2928: UI: Export Query Results to CSV on DB Catalog

### Description

I want to be able to export query results from the Query tab of the Database Catalog to a CSV file. We have a `ToCsvReactor` that can be utilized. Ideally this button is displayed next to the collapse button in the Query Results row. 

Image

### Relevant Feature or Code

_No response_

### Requirements (what)

_No response_

kellogg9:
## Description
Added button to download query results to CSV.

## Changes Made
Passes SQLQuery() pixel call to query-results-panel.tsx , pipes to ToCsv pixel call, and then downloads

## How to Test
- Head to Databases tab, click on one, and go to Query tab

- Write a SQL statement on your Database, click Run Query button to see that Query Results field is populated

- Click Download button (File Download icon)

- Check to see if correct CSV table was downloaded


ppatel9703:
We have reactors to handle exports to excel and csv -- we should use those to get them hooked to the FE

tevanburen:
Consider having a tooltip that says something to the effect of "Re-run query and download results" to address the edge case we discussed (db is updated in between viewing and downloading)

PR #2966, opened 2026-04-02
Use theme.store for feature flags, and code cleanup

### LINKED ISSUE #2961: Hide branching (FDA request)
Have the rewrite button be controlled by an env flag

tevanburen:
## Description
Store .env variables in the theme map too
Add enableRewrite to control whether users can resend messages

## Changes Made
- Random lint stuff (I'm not sure why my ts is being aggressive today?)
- Add featureFlags to the theme map, and add in .env variables
- Where the .env variables were used, instead use theme map (see notes)
- Add a enableRewrite flag, hide that option when false
- In global-nav.tsx, optimistically delete rooms (show as deleted before BE returns, then repopulate if call fails)
- Remove _error field from chat.store because it was unused

## How to Test
1. Get on this branch
2. In your .env.local, set:
```
VITE_ENABLE_MODEL_SELECT=false
VITE_ENABLE_AGENT=false
VITE_ENABLE_SUGGESTIONS=false
VITE_ENABLE_PLAN=false
VITE_ENABLE_REWRITE=true
```
3. Verify that model selection, agent page, suggestions, plan mode, and ability to rewrite are gone
4. Add the following to your theme map:
```
{
  ...etc,
  "playground": {
    ...etc,
    "featureFlags": {
      "enablePlan": true,
      "enableAgent": true,
      "enableSuggestions": true,
      "enableModelSelect": true,
      "enableRewrite": true
    }
  }
}
```
5. Verify that model selection, agent page, suggestions, plan mode, and ability to rewrite are back

## Notes
The values of the .env variables are essentially the defaults. If you add values to the theme map, the theme map will override .env variables. For example, if I put enableRewrite=false in .env, and theme map is empty, then rewrite will be disabled. If I then add enableRewrite: true to the theme map, then rewrite will be enabled


ckelly17 (approved):
Reviewed with Thomas. Functionality is as described

kzsb03 (approved):

PR #3004, opened 2026-04-08
feat(playground): image gen models and settings

### LINKED ISSUE #2983: Image Parts Rendering

### Description

See the following issues:
1. If you have an image part as part of the response message, and the image was downloaded to the room folder, when I click on the image icon it errors trying to open it. I can open it fine from the file explorer. See attached video.
2. When I log out and log in, the image renderings don't persist. See attached photo
3. Feedback doesn't seem to work for those response messages.

### Relevant Feature or Code

Image

https://github.com/user-attachments/assets/d28fdb57-c71a-45fe-82a8-aac75e04e11d

Image

### Requirements (what)

Use this branch to test on the backend (if not merged): https://github.com/SEMOSS/Semoss/pull/2291

kellogg9:
Added frontend handling to playground for image generation models.

List of features:
* Updated model dropdown to include 'image-generation' tagged models
* Better image rendering in room chats
* Image Generation field in Room Settings for users to adjust orientation, size, and seed parameter values
* UI Updates to Text Generation field in Room Settings
* Toggleable through feature flags

On this branch, the feature is currently set to false (i.e. "enableImageGeneration": false). To see image generation models and settings in playground, create an admin theme and change this flag to true.


libs/shared/src/types.ts
```diff
@@ -111,6 +111,11 @@ export interface ThemeMap {
 			model?: Engine;
 			temperature?: number;
 			tokenLength?: number;
+			numOfImages?: number;
+			imageHeight?: number;
+			imageWidth?: number;
+			cfgScale?: number;
+			seed?: number;
 		};
```
tevanburen, line 138: also add to the default vals stored in admin-theme-page.tsx

libs/i18n/src/resources/locales/ja/common.json
tevanburen, line 1: havent checked these yet but can do as very last step

packages/playground/src/components/message/response-message.tsx
```diff
@@ -176,6 +204,15 @@ export const ResponseMessage: React.FC<ResponseMessageProps> = observer(
 								/>
 							);
 						} else if (p.type === "MEDIA") {
+							if (hasOnlyMedia) {
```
tevanburen: should this not be also for messages with media and text?

packages/playground/src/components/room/room-input.tsx
```diff
@@ -745,7 +747,14 @@ export const RoomInput: React.FC<RoomInputProps> = observer(
 										value={model?.app_id || ""}
 										engineTypes={["MODEL"]}
 										metaFilters={[
-											{ tag: "text-generation" },
+											{
```
tevanburen: may be less confusing to just do 
```
tag: isImageGenEnabled ? ["text-generation", "image-generation"] : ["text-generation"]
```

packages/playground/src/components/room/room-options-form.tsx
```diff
@@ -61,6 +65,44 @@ export const RoomOptionsForm: React.FC<RoomOptionsFormProps> = observer(
 			isOpen: false,
 		});
 
+		const deriveImagePreset = () => {
```
tevanburen: ok this stuff is a bit ugly, we should find a better way to map to height and width and vice versa

packages/playground/src/components/room/room-options-form.tsx
```diff
@@ -447,6 +502,242 @@ export const RoomOptionsForm: React.FC<RoomOptionsFormProps> = observer(
 							</Field>
 						</FieldGroup>
 					</FieldSet>
+					{isImageGenEnabled && (
```
tevanburen: honestly let's just pull this whole section into a new component

packages/playground/src/stores/message/utility.ts
```diff
@@ -1,9 +1,32 @@
 import type { RoomStore } from "@/stores";
-import type { PixelMessage } from "@/types";
+import type { Engine, PixelMessage } from "@/types";
 import { InputMessageStore } from "./input-message.store";
 import { PlanMessageStore } from "./plan-message.store";
 import { ResponseMessageStore } from "./response-message.store";
 
+/**
```
tevanburen, line 7: We should find a better way to do this

packages/playground/src/stores/room/room.store.ts
```diff
@@ -713,9 +755,24 @@ export class RoomStore {
...
+				imageHeight: _imageHeight,
+				imageWidth: _imageWidth,
+				cfgScale: _cfgScale,
+				seed: _seed,
+				...baseOptions
+			} = options;
+
 			const optionsToSave = {
-				...options,
+				...(isImageModel ? options : baseOptions),
```
tevanburen: dosh should update so that this isn't necessary imo

tevanburen: also this doesn't work for models with both images and text

packages/playground/src/constants.ts
```diff
@@ -3,6 +3,28 @@ export const MODEL_KEY = "SMSS-SELECTED-MODEL-V3";
...
+	"small-portrait": { height: 384, width: 320 },
+	"small-landscape": { height: 320, width: 448 },
+	"medium-square": { height: 448, width: 448 },
+	"medium-portrait": { height: 512, width: 384 },
+	"medium-landscape": { height: 384, width: 512 },
+	"large-square": { height: 1024, width: 1024 },
+	"large-portrait": { height: 1024, width: 960 },
+	"large-landscape": { height: 448, width: 768 },
+};
+
```
tevanburen, line 36: tweak to make more distinct

tevanburen (changes requested):

libs/i18n/src/resources/locales/en/common.json
```diff
@@ -16,14 +16,17 @@
...
+		"imageGenHeader": "Image Generation Settings"
 	},
 	"placeholders": {
 		"enterName": "Enter Name",
 		"enterDescription": "Enter Description",
 		"enterInstructions": "Enter Instructions",
 		"updateInstructions": "Update Instructions",
-		"updateTokenLength": "Update token length",
+		"updateMaxToken": "Update max tokens",
+		"updateImageSeed": "1",
```
tevanburen: "updateImageSeed": "1" ?

should probably just have this not be an i18n key i guess

packages/playground/src/components/message/response-message.tsx
```diff
@@ -176,6 +177,30 @@ export const ResponseMessage: React.FC<ResponseMessageProps> = observer(
 			{ value: "pdf", label: "PDF Document", extension: ".pdf" },
 		];
 
+		const copyImage = async () => {
+			const mediaPart = message.parts.find((p) => p.type === "MEDIA");
```
tevanburen, line 224: do we think there's a way to copy multiple images if we have them?

tevanburen: Claude says fetch() round-trip is unnecessary — converting a base64 string to a Blob via fetch(data:…) works but is roundabout. You can do it directly:

- const bytes = atob(mediaPart.mediaInfo.base64Data);
- const arr = new Uint8Array(bytes.length).map((_, i) => bytes.charCodeAt(i));
- const blob = new Blob([arr], { type: mimeType });

The fetch approach spawns a micro-network request and is slower on large images.

packages/playground/src/constants.ts
```diff
@@ -1,8 +1,39 @@
 export const MODEL_KEY = "SMSS-SELECTED-MODEL-V3";
 
-export const TOKEN_LENGTH: number = undefined as unknown as number; // token length is determined at runtime based on the model, so we initialize it as undefined and set it later
+export const MAX_TOKENS: number = undefined as unknown as number; // token length is determined at runtime based on the model, so we initialize it as undefined and set it later
 export const TEMPERATURE = 0.3;
 
+export const IMAGE_HEIGHT = 1024;
+export const IMAGE_WIDTH = 1024;
+export const TEXT_GENERATION_FLAG = false;
+export const IMAGE_GENERATION_FLAG = false;
```
tevanburen: these names are misleading to me, from the name TEXT_GENERATION_FLAG I would think the value would be the key itself, like `export const TEXT_GENERATION_FLAG = "text-generation"`

Recommend something like TEXT_GENERATION_DEFAULT_ENABLED etc

packages/playground/src/stores/message/response-message.store.ts
```diff
@@ -233,7 +234,7 @@ paramValues=[${JSON.stringify({
 								});
 							}
 						} else if (chunk.stream_type === "thinking") {
-							if (chunk.data.thinking) {
+							if (!isImageModel && chunk.data.thinking) {
```
tevanburen: why?

packages/playground/src/stores/message/response-message.store.ts
```diff
@@ -343,6 +344,7 @@ paramValues=[${JSON.stringify({
 		const room = this.room;
 
 		try {
+			// idk why failing for image response
```
tevanburen: this still a problem?

tevanburen (commented):
I think we still want the image gen model chip for now too

tevanburen:
Oh and I think we need helper text on the seed thing telling users the bounds

packages/playground/src/stores/message/response-message.store.ts
kellogg9: Don't think the image model's was testing with return thinking chunks but ig can remove in case some do

kellogg9:
> I think we still want the image gen model chip for now too

resolved

kellogg9:
> Oh and I think we need helper text on the seed thing telling users the bounds

resolved

PR #3109, opened 2026-04-22
feat(playground): use temp insights for runMCPTool

tevanburen:
## Description

Use temp insights for RunMCPTool for reasons determined by @rithvik-doshi and @agmontem-devel 

## How to Test

1. Prompt playground for tools (both auto-execute and ask with default UI)
2. Verify all tools ran have SetRoomForInsight and are on their own insights


rithvik-doshi (approved):

tevanburen:
@rithvik-doshi are we good to close this out?

tevanburen:
Doshi says BE covered

PR #3121, opened 2026-04-23
Tvb file drop 2

### LINKED ISSUE #2993: Enhance file drop state
When users drag files into playground, have it be better at receiving

Maher:
> how hard on playground would it be via a theme, that the add document gives a dropzone popup instead of directly opening the filesystem so we can some themed text for a notice to the user like "Please ensure you have the appropriate approvals for using any 3rd party materials with AI technologies" or w/e text OGC wants 

Henry: 
> when dragging in attachments it has to be in that first row

tevanburen:
## Description
Replaces the inline drag-and-drop handling in the chat input with a modal-based attachment flow. Previously, files were managed with local state inside RoomInput, with drag events handled on the input container itself. Now, a FileDragProvider context owns all file state, and dragging over the input area (or clicking the upload menu item) opens a dedicated "Add Documents" dialog with a drop zone and scrollable file preview strip. Once confirmed, files appear in a persistent thumbnail strip inside the chat input and are cleared automatically on send (or restored on error).

## Changes Made
**Context (file-drag-context.tsx)**
- New FileDragProvider and useFileDrag hook own all file state (files, isDragging, shouldStayOpen) and the shared fileInputRef
- Window-level dragover/drop listeners handle file capture across the page; a containerRef bounds check prevents the modal from triggering when a fixed/overlay element is above the drop target
- A 300ms abandon timer cancels isDragging if the drag leaves the window without dropping
- FILE_DRAG_ATTR data attribute marks the portaled dialog so boundary checks identify it correctly

**New Components**
- FileDragOverlay (file-drag-overlay.tsx): Dialog-based attachment UI with a fixed-height drop zone, file strip that hides when empty, "Browse" and "Done" buttons, and optional theme disclaimer rendered in DialogDescription
- FilePreviewGrid (file-preview-grid.tsx): Horizontally scrollable thumbnail strip extracted from RoomInput; image preview URLs are built synchronously via useMemo with a separate effect to revoke on map replacement or unmount; renders type-specific Lucide icons for non-image files

**Room Input (room-input.tsx, room-input-menu-upload.tsx, room-content.tsx)**
- Removed local isDragging/files state, fileRef, image preview memo, and all onDrop/onDragOver/onDragLeave handlers from RoomInput
- File state now sourced from useFileDrag; addFiles/removeFile/clearFiles replace direct setFiles calls
- RoomInputMenuUpload no longer accepts a fileRef prop — it calls setShouldStayOpen(true) to open the modal instead
- fileRef removed from MenuComponent prop interface and all call sites

**Pages (room-page.tsx, new-room-page.tsx)**
- RoomContent and the new-room input are each wrapped in FileDragProvider + FileDragOverlay; the provider is scoped to the panel, not the whole page

**i18n**
- New fileDrag namespace (title, modalTitle, browse, done) added to all six locale files (en, ar, es, fr, hi, ja)

**Theme**
- ThemeMap gains an optional fileDragDisclaimer field; RootStore merges it from the theme config and exposes it to the overlay's DialogDescription

## How to Test
1. Drag to open — drag a file over the chat input area; modal should open with the drop zone highlighted
2. Drop in modal — drop the file; it should appear in the file strip, drop zone should remain visible below
3. Drop on backdrop — drag a file and drop it on the modal backdrop (not the drop zone); file should still be captured
4. Drag outside territory — drag a file over a different section of the app (sidebar, header); modal should not open
5. Drag away mid-drag — drag over the input, then drag out of the provider area without dropping; modal should close
6. Click drop zone — with modal open, click the dashed area; file explorer should open
7. Browse button — click "Browse" at the bottom; file explorer should open
8. Upload menu item — click the paperclip/upload button in the input menu; modal should open without a drag
9. Multiple files — add enough files to require horizontal scrolling in the strip; strip should scroll, modal width should stay fixed
10. Remove file — hover a thumbnail, click the X; file should be removed
11. Done button — click Done; modal closes, files appear as thumbnails in the chat input strip
12. Send clears files — send a message with files attached; strip should be empty afterward
13. New room page — verify the same drag/modal behavior works on the new-room (pre-conversation) page
14. Disclaimer — if fileDragDisclaimer is set in the theme config, verify it appears as a subtitle under "Add Documents"

## Notes
- **Two independent FileDragProvider scopes:** one in RoomPage, one in NewRoomPage. They do not share state intentionally — each panel manages its own attachment queue.
- **Drop boundary logic is layered:** the React onDragOver on the container div guards initial activation (bails if a position: fixed ancestor is detected); the window listener then handles the actual drop anywhere on the page once active. This avoids the modal backdrop swallowing drop events.
- **shouldStayOpen vs isDragging:** the modal stays open if either flag is true. shouldStayOpen is set to false on send and cleared on close, so files don't leak between messages.
- **Paste-to-attach still works:** onPaste in RoomInput calls addFiles into the shared context; files will appear in the strip without opening the modal.


kellogg9:
LGTM

PR #3127, opened 2026-04-24
Fix mobile issue 1033

yifan-z0698:
## Description
Fix the Select Agent submenu overflowing beyond the viewport on mobile devices by rendering it as a bottom drawer instead of a nested `DropdownMenuSub`.

## Changes Made
- Detect mobile viewport using `useIsMobile()` hook
- On mobile: render a `DropdownMenuItem` that opens a `Drawer` (bottom sheet) containing the agent picker
- On desktop: keep the existing `DropdownMenuSub` / `DropdownMenuSubContent` behavior unchanged
- Extract the shared agent picker content (search input, command list, workspace items) into a `pickerContent` variable to avoid duplication between mobile and desktop paths
- Add `setIsOpen(false)` on agent selection to properly close the picker after choosing an agent

## How to Test
1. Open the playground using Chrome extension ```Mobile simulator```
2. Click the input menu (plus/options button) in the chat input area
3. Tap "Select Agent" — it should open as a bottom drawer sliding up from the bottom of the screen
4. Search for and select an agent — the drawer should close and the agent should be applied

## Notes
- This change does not affect the desktop UI’s appearance or the desktop user experience.
- Only `room-input-menu-workspace.tsx` is modified in this PR
- New UI imports added: `Drawer`, `DrawerContent`, `DrawerHeader`, `DrawerTitle`, `DropdownMenuItem`, `useIsMobile`
- Screenshot:
image


packages/playground/src/components/room/room-input-menu-workspace.tsx
```diff
@@ -87,103 +94,135 @@ export const RoomInputMenuWorkspace: React.FC<RoomInputMenuWorkspaceProps> = ({
 		},
 	});
 
+	const triggerLabel = workspace
+		? workspace.name
+		: t("menuWorkspace.selectAgent");
+
+	const pickerContent = (
```
kzsb03, line 102: why is this moved out?

yifan-z0698: It is just deduplication. The same label logic (workspace ? workspace.name : t("menuWorkspace.selectAgent")) is now used in two places — the mobile DropdownMenuItem trigger (line 191) and the desktop DropdownMenuSubTrigger (line 217). Extracting it into triggerLabel avoids repeating the same ternary expression in both branches.

packages/playground/src/components/room/room-input-menu-workspace.tsx
```diff
@@ -49,6 +55,7 @@ export const RoomInputMenuWorkspace: React.FC<RoomInputMenuWorkspaceProps> = ({
 	const { t } = useTranslation("room");
 	const { root } = useRoot();
 	const [search, setSearch] = React.useState("");
```
tevanburen: can just use useState

packages/playground/src/components/room/room-input-menu-workspace.tsx
```diff
@@ -87,103 +94,135 @@ export const RoomInputMenuWorkspace: React.FC<RoomInputMenuWorkspaceProps> = ({
...
+		? workspace.name
+		: t("menuWorkspace.selectAgent");
+
+	const pickerContent = (
+		<Command shouldFilter={false} className="w-full">
+			<CommandInput
+				placeholder={t("menuWorkspace.searchPlaceholder")}
+				value={search}
+				onValueChange={setSearch}
+				autoFocus
```
tevanburen: this is opening the keyboard on mobile which is a bit weird, so can do `autoFocus={!isMobile}`

yifan-z0698: Good catch — fixed.

packages/playground/src/components/room/room-input-menu-workspace.tsx
yifan-z0698: Fixed

packages/playground/src/components/room/room-input-menu-workspace.tsx
kzsb03: I'm referring to the picker content

yifan-z0698: The pickerContent is also used in two place line 207 and line 226 for mobile and non-mobile UI. So the reason is also deduplication.

kzsb03 (approved):

PR #3199, opened 2026-05-07
Designate platform tools through the admin theme map

rithvik-doshi:
## Platform tools — frontend

Adds frontend support for the backend's new platform tool category. The LLM is given platform tools alongside per-project MCP tools.

### Changes

- **Types** — adds `playground.platformTools: string[]` to `ThemeMap` (`libs/shared/src/types.ts`) and an `_meta.SMSS_IS_PLATFORM_TOOL?: boolean` flag on `MCPTool` / `PixelMessageToolCallPart` (`packages/playground/src/types.d.ts`); clarifies that the existing `defaultTools` is for MCP servers only.
- **Theme stores** — initializes and merges `platformTools` across theme inheritance in the root store and the admin theme page's `EMPTY_PLAYGROUND`.
- **No-project handling** — `tools-default-view.tsx` and `tools-view.tsx` skip the `GetMCPTools` / ProjectInfo fetch when no app is bound, since platform tools have no owning project; renders the default view directly.
- **Room creation** — `new-room-page.tsx` no longer spreads `defaultTools` into the new-room `mcp` option (platform tools are discovered server-side from the active theme, not enumerated on the client).
- **Tool error surface** — `response-message.store.ts` now reads `operationType` from the Pixel return and flags `toolError = true` when it contains `"ERROR"`, so platform tool failures render as tool errors instead of silent successes.

### Pairs with

Backend PR `platform-file-utils` (Semoss)


packages/playground/src/pages/new-room-page.tsx
```diff
@@ -401,7 +403,9 @@ export const NewRoomPage = observer(() => {
 				instructions: "",
 				temperature: root.theme.defaultRoomSettings?.temperature,
 				tokenLength: root.theme.defaultRoomSettings?.tokenLength,
-				mcp: [...(root.theme.defaultTools || [])], // Remove workspace MCPs
+				mcp: (root.theme.defaultTools || []).filter(
+					(t) => t && typeof t === "object",
+				), // Remove workspace MCPs; platform default tools are discovered by the backend
 			});
```
tevanburen, line 397: why?

packages/playground/src/constants.ts
```diff
@@ -6,6 +6,8 @@ export const TEMPERATURE = 0.3;
 export const MCP_EXECUTION_AUTO = "auto";
 export const MCP_EXECUTION_ASK = "ask";
 
+export const PLATFORM_TOOL_PREFIX = "platform__";
```
tevanburen: should pass whether is platform in meta map

PR #3229, opened 2026-05-19
feat: a new filter data cell test has been implemented

boredcod:
## Description
Added a frontend unit test for the filter data cell

## How to Test

cd into /libs/renderer/src/testing/cell-defaults and run pnpm test filter-data-cell.spec.tsx


memisrose (approved):
Image

PR #3235, opened 2026-05-21
feat(playground): redesign mcp selector and move engine assets to libs/shared

themaherkhalil:


themaherkhalil:
Replacing with https://github.com/SEMOSS/semoss-ui/pull/3226

PR #3237, opened 2026-05-21
fix: render input params a bit better

rithvik-doshi:
## Description
Handle the json passed for input paramters in the tools view


ekim1707 (approved):
these changes look good to me

kzsb03 (approved):

PR #3260, opened 2026-05-26
feat: fix for chats showing in wrong chat history time of the day/week

SunTheDev:
Description

When a user's local timezone is behind UTC (e.g. EDT = UTC-4), chats created after 8 PM local time were incorrectly bucketed under "Yesterday" in the sidebar instead of "Today." This is because the server's systemDate is stored in UTC without a timezone indicator, and the frontend was parsing it as local time — making it one calendar day ahead of the chat timestamps, which were already being parsed correctly as UTC.

Changes Made

In global-nav.tsx, appended Z to system.config.systemDate when constructing the systemDate dayjs object:
const systemDate = dayjs(`${system.config.systemDate}Z`);
This makes systemDate consistent with how DATE_CREATED is already parsed — both are now treated as UTC and converted to local time before comparison.

Steps to Reproduce / Test the Behavior

Be in a timezone that is behind UTC (e.g. EDT, CDT, PDT).
After UTC midnight (e.g. after 8 PM EDT), create a new chat or look at chats created earlier that day.
Open the chat history sidebar and observe that those chats appear under "Yesterday" instead of "Today."
After the fix, the same chats correctly appear under "Today."
Expected Outcomes
Chats created on the user's current local calendar day are always bucketed under "Today," regardless of the UTC date.
The displayed timestamp (e.g. 5/21/2026, 8:05 PM) is unchanged — only the bucket label is affected.
No behavioral change for users in UTC or timezones ahead of UTC.

Notes

systemDate is only used for bucket label comparisons in global-nav.tsx — this change has no effect on any other functionality.
The bug only manifests after UTC midnight (the exact cutoff depends on the user's UTC offset), which is why it may not be reproducible earlier in the day.
DATE_CREATED already had Z appended correctly — this change simply applies the same treatment to systemDate for consistency.


ekim1707 (approved):
Looks good to me codewise

kzsb03 (approved):

PR #3261, opened 2026-05-26
Slash command overhaul

### LINKED ISSUE #3239: Overhaul slash command

tevanburen:
## Description
Replaces the old slash menu (a searchable, backend-fetched list of MCP servers) with a static command palette that opens Playground's existing MCP overlay panels and triggers in-conversation actions. Selected commands appear as styled inline chips in the editor input; chip labels are included as plain text when the message is sent.

## Changes Made

**mention-plugin.tsx**
- Added full keyboard handling: arrow up/down navigate items, Enter accepts, Tab completes, Escape closes without clearing typed text
- Menu position is locked to the trigger character's location and flips above the cursor when near the bottom of the viewport
- Added `addNode(factory)` API (factory called inside `editor.update()` to avoid the "no active editor" error when triggered by click)
- Added `onAccept` and `onTabComplete` callbacks; exposes `query`, `selectedIndex`, `setItemCount`, `setSelectedIndex` to the menu component

**slash-command-node.tsx** *(new)*
- Inline Lexical `DecoratorNode` that renders a styled chip; `isKeyboardSelectable: false` so arrow keys skip over it cleanly
- `getLabel()` used during text extraction on send so chip content is included in the message

**room-input-menu-slash.tsx** *(replaced)*
- Previous implementation fetched MCP servers from the backend with infinite scroll; new implementation is a static list of predefined commands
- `SlashCommand` type supports `hiddenInMenu` (hidden until first character typed) and `noChip` (fires immediately, no chip inserted)
- `buildSlashCommands(onOpenMcpOverlay, onCompact)` — commands: `/knowledge`, `/toolbox`, `/agent`, `/compact`, plus hidden aliases `/mcp` and `/workspace`
- `filterSlashCommands` is exported and shared between the menu and `room-input.tsx` so selection index always aligns with the visible list

**room-input.tsx**
- Registered `SlashCommandNode` in the Lexical composer; empty-state check accounts for chips (chip alone enables the send button)
- Text extraction on send walks the tree manually to include chip labels; `$nodesOfType` (deprecated) replaced with tree walk
- `noChip` commands execute immediately and close the menu without inserting a chip
- `excludeCommandIds` prop lets call sites suppress specific commands; MCP overlay refocuses the editor on close

**room-content.tsx / new-room-page.tsx**
- Removed now-dead `handleToolAdd` / `onMcpSelect`
- `room-content`: `excludeCommandIds={["agent", "workspace"]}` — agent selection not applicable in existing rooms
- `new-room-page`: `excludeCommandIds={["compact"]}` — no conversation history to compact before a room exists

## How to Test
1. In an existing room, type `/` in the input — a command menu should appear anchored to the cursor.
2. Arrow up/down to navigate; the highlighted item should scroll into view. Press Tab to autocomplete the query text.
3. Select `/knowledge`, `/toolbox` — the MCP overlay should open on the correct tab, and a chip should appear in the input.
4. Select `/agent` — command should not appear in the existing room menu (excluded).
5. Type `/compact` and select it — no chip should be inserted; a compaction toast should fire.
6. On the new room page, confirm `/compact` does not appear in the menu and `/agent` does.
7. With a chip in the input, press the arrow keys — cursor should skip over the chip without disappearing.
8. Press Escape while the menu is open — menu should close, typed text (e.g. `/kno`) should remain in the editor.
9. Send a message containing a chip — the chip's label (e.g. `/knowledge`) should appear as plain text in the submitted message.

## Notes
The old slash menu fetched and displayed MCP servers directly; that lookup now lives entirely in the MCP overlay panels that the commands open. The `MentionPlugin` changes are generic — nothing in the plugin is Playground-specific, so it can be reused with a different trigger or command set.

The Playground system prompt should be updated to describe the available slash commands so the agent doesn't make assumptions about them:
```
You are an AI assistant embedded in Playground, the Semoss platform's chat interface, deployed for internal Deloitte use. You assist with technical and analytical work including software development, data analysis, research, and writing. Assume the user has professional expertise in their domain — don't over-explain fundamentals unless asked. Write in complete, natural sentences; avoid fragments or dramatic one-liners. Stay focused on what was asked and don't volunteer unrequested suggestions or caveats. If the user is wrong or missing something, say so directly rather than agreeing to be agreeable. Messages may include slash commands. Any tools, knowledge, or other resources available to you are the attached context — when a slash command is present, acknowledge it and surface what's available so the user knows you see it. Don't invent capabilities or context that aren't present.
```


kellogg9:
LGTM

packages/playground/src/components/room/room-input.tsx
tevanburen, line 1: The majority of the changes down in the return statement are whitespace only

kzsb03 (approved):

kzsb03 (approved):

PR #3300, opened 2026-06-09
playground typewriter updates

### LINKED ISSUE #3275: FDA typewriter fixes
- Continue typewriter when the user jumps off the page (rn it pauses I think?)
- Jump to end when streaming is done

kellogg9:
## Description
Refactors the chat response rendering pipeline so a message's parts — thinking, text, tools, media — reveal as a single sequential, animated queue instead of all at once. A message plays from the beginning the first time it streams in front of you and jumps straight to the latest content when you return to one already in progress, with the same playback logic driving both the part/chunk reveal order and the typewriter.

## Changes Made

**`use-active-index.ts` (new):** Sequential reveal queue hook. Tracks the active index, hands each item a stable `onComplete` to advance the queue, holds the last item while streaming (more content may arrive), and bubbles completion up to a parent queue. Takes an optional `startAtLatest` to seed at the frontier on a return view.

**`parse-chunks.ts` (new):** Parses streamed text into an ordered array of `md` and `html` chunks by scanning for ` ```html ``` ` fences, with stable `key`s so React reuses subcomponents across re-parses.

**`use-markdown-typewriter.ts`:** Adds a `startAtLatest` option that seeds the reveal baseline at the content present on mount, so on a return only net-new tokens animate (jump to latest, no replay). First view still animates from 0.

**`response-message.tsx`:** Message-level orchestrator. Runs the queue over `message.parts` so each part waits for the one above to finish; text parts type, while thinking/media/tool parts snap to their final state but still wait their turn (advanced via an effect). Captures a single `isFirstView` decision at mount (the message had no streamed content yet) and propagates it down. Also gates the "complete tools to proceed" prompt behind `!isThinking` so it only shows once streaming ends.

**`response-message-text.tsx`:** Parses a text part into chunks and runs a nested chunk queue, reporting its own completion upward when the last chunk catches up. Memoizes `parseChunks` on the text and seeds at the latest chunk on a return view.

**`response-message-text-md.tsx` (new):** Animates a single markdown chunk; typewriter is driven from an effect (so the parent-queue update happens post-commit, not during render). Baseline comes from `isFirstView`. `"done"` renders full content directly.

**`response-message-text-html.tsx` (new):** Renders a single HTML chunk — loading state while the fence is open, full preview once the parent marks it done.

**`response-message-thinking.tsx`:** Driven by a `status` prop instead of `isLast`/`isStreaming`; uses the same first-view/return typewriter baseline. Types the markdown component overrides and fixes the `blockquote` override (was rendering a Lucide icon).

**`create-markdown-components.tsx`:** Drops the inline ` ```html ``` ` fence handling (HTML is now its own chunk type) and the now-unused translation hook.

**`room-content.tsx`:** Fixes a scroll timing issue where the smooth-scroll-to-bottom fired before the typewriter finished dumping remaining content after streaming ended, landing the scroll short.

**`room.store.ts`:** Streaming poll interval 300ms → 500ms (the typewriter keeps output looking live between polls).

## How to Test
1. Send a prompt that produces a long markdown response — text animates from the very beginning (not mid-sentence) and snaps to full content the moment streaming ends.
2. Send "Explain what a pie chart is in one sentence, then give me a pie chart as an HTML page, then tell me one fun fact about charts" — all three chunks animate in order without hanging.
3. Send "Give me two separate HTML pages — one red background, one blue background" — both HTML blocks render sequentially.
4. While a long response is streaming in an existing room, switch to another room and back — content is caught up to the latest with no replay, and continues forward from there.
5. While streaming, switch browser tabs and return — content snaps to latest with no re-animation.
6. Open an old conversation — all parts render immediately with no animation.
7. Send a prompt that triggers tools — the "complete tools to proceed" prompt only appears after the message finishes streaming, not during.


packages/playground/src/components/message/response-message-text/parse-chunks.ts
tevanburen, line 1: I think this is clean

packages/playground/src/components/message/response-message-text.tsx
tevanburen, line 1: I think this is clean but see comments in the md file, I think some of that logic should be moved in here

packages/playground/src/components/message/response-message-text/response-message-text-md.tsx
```diff
@@ -0,0 +1,208 @@
...
+
+			// Per-activation guards — reset together whenever this chunk becomes active
+			const onCompleteCalledRef = useRef(false);
+			const typewriterStartedRef = useRef(false);
+			useEffect(() => {
+				if (isActive) {
+					onCompleteCalledRef.current = false;
+					typewriterStartedRef.current = false;
+				}
+			}, [isActive]);
```
tevanburen: I'm not sure I see the necessity of these refs. I think that trying to guard the child from calling onComplete is kinda logic that should be in the parent, i.e. response-message-text should make sure that onComplete calls from its children are valid before advancing its activeIndex

packages/playground/src/components/message/response-message-text/response-message-text-md.tsx
```diff
@@ -0,0 +1,208 @@
...
+
+			const fireOnComplete = useCallback(() => {
+				if (!onCompleteCalledRef.current) {
+					onCompleteCalledRef.current = true;
+					onComplete();
+				}
+			}, [onComplete]);
```
tevanburen: onComplete itself is re-rendered every time response-message-text is, so this callback is unnecessary

packages/playground/src/components/message/response-message-text/response-message-text-md.tsx
```diff
@@ -0,0 +1,208 @@
...
+			}, [isActive, newContent.length, isFinalized]);
+
+			// Skip animation for done chunks (e.g. when isLast prop changes)
+			useEffect(() => {
+				if (isDone) {
+					typewriter.skipToEnd();
+				}
+			}, [isDone, typewriter.skipToEnd]);
+
+			// When streaming ends (isFinalized becomes true) and the typewriter is
```
tevanburen: this is a bit inconsistent logic I think - isFinalized is whether that specific chunk is done streaming, but not whether the message itself is

packages/playground/src/components/message/response-message-text/response-message-text-md.tsx
```diff
@@ -0,0 +1,208 @@
...
+				// stops isRunning, which will trigger the caught-up effect below.
+				typewriter.skipToEnd();
+			}, [isFinalized]);
+
+			// Fire onComplete when typewriter has caught up AND chunk is finalized.
+			// Only fires when there is actual new content to animate — the "no new
+			// content" path is handled by the start effect above (fireOnComplete when
+			// isFinalized and newContent is empty), preventing a premature complete
+			// before the typewriter has had a chance to start.
+			useEffect(() => {
```
tevanburen: this kinda is the same logic as above tho no?

packages/playground/src/components/message/response-message-text/response-message-text-md.tsx
```diff
@@ -0,0 +1,208 @@
...
+				}
+			}, [
+				isActive,
+				typewriter.isTyping,
+				typewriter.rendered.length,
+				newContent.length,
+				isFinalized,
+				fireOnComplete,
+			]);
+
```
tevanburen: in general, a million useEffects can be a code smell

packages/playground/src/components/message/response-message-text/response-message-text-html.tsx
tevanburen, line 1: I think this file probably has the same comments as those in the md one

packages/playground/src/components/message/response-message-text/response-message-text-md.tsx
kellogg9: Good point, resolved

packages/playground/src/components/message/response-message-text/response-message-text-md.tsx
kellogg9: updated to be isStreaming

packages/playground/src/components/message/response-message-text/response-message-text-md.tsx
kellogg9: it decreases the delay between when the LLM has finished streaming and when the remaining buffered text is fully rendered. we can remove it though, I think the delay is negligible.

packages/playground/src/components/message/response-message-text/response-message-text-md.tsx
kellogg9: agreed - is your solution moving some of the state to the typewriter hook (i.e. having it accept isFinaliized and handling skipToEnd)? its purely animation atm

packages/playground/src/components/message/response-message-text/response-message-text-md.tsx
```diff
@@ -0,0 +1,201 @@
...
+						root.theme.allowedUrlPrefixes?.some((prefix) =>
+							url.startsWith(prefix),
+						)
+					)
+						return url;
+					if (/^(https?:|mailto:|#)/.test(url)) return url;
+					return "";
+				},
+				[root.theme.allowedUrlPrefixes],
+			);
```
tevanburen: This can be a normal method instead of a useCallback

packages/playground/src/hooks/use-markdown-typewriter.ts
```diff
@@ -139,6 +139,31 @@ export function useMarkdownTypewriter(
 		return content.slice(0, renderedLength);
 	}, [content, renderedLength]);
 
+	// Tab-away handling:
+	// - hidden: snapshot renderedLength at departure (locks in what was shown)
+	// - visible: snap renderedLength forward to absorb everything that streamed
+	//   in while away, then reset timing so only net-new tokens get animated.
+	useEffect(() => {
+		const handleVisibilityChange = () => {
+			if (document.visibilityState === "hidden") {
```
tevanburen: dunno if this case is needed, because we can just rely on the else if to handle this

packages/playground/src/hooks/use-markdown-typewriter.ts
tevanburen, line 194: I think this flow is nor no longer necessary, because you split out code chunks in your parseChunks right?

packages/playground/src/hooks/use-markdown-typewriter.ts
```diff
@@ -215,7 +240,7 @@ export function useMarkdownTypewriter(
 
 	return {
 		rendered,
-		isTyping: isRunning && renderedLength < content.length,
+		isTyping: isRunning && renderedLength < contentRef.current.length,
```
tevanburen: Could you explain this line change?

packages/playground/src/components/message/response-message-text/response-message-text-md.tsx
```diff
@@ -0,0 +1,201 @@
...
+
+			// ── Effects ───────────────────────────────────────────────────────────────
+
+			// Track whether the typewriter has been started for the current activation.
+			// Reset whenever isActive flips on so a re-activation starts fresh.
+			const typewriterStartedRef = useRef(false);
```
tevanburen: I don't think we need this ref; I think we can just call `typewriter.start()` in the useEffect below when `newContent.length > 0`, without the ref check, because as the comment notes, that method is stable

packages/playground/src/components/message/response-message-text/response-message-text-md.tsx
```diff
@@ -0,0 +1,201 @@
...
+			useEffect(() => {
+				if (isActive) {
+					typewriterStartedRef.current = false;
+				}
+			}, [isActive]);
+
+			// biome-ignore lint/correctness/useExhaustiveDependencies: typewriter.start is stable (useCallback with empty deps) — omitting it is safe and prevents spurious re-triggers
```
tevanburen: Since the method is stable we can just add it as a dep anyway

packages/playground/src/components/message/response-message-text/response-message-text-md.tsx
```diff
@@ -0,0 +1,201 @@
...
+					onComplete();
+				}
+			}, [isActive, newContent.length, isFinalized]);
+
+			// Skip animation for done chunks (e.g. when isLast prop changes)
+			useEffect(() => {
+				if (isDone) {
+					typewriter.skipToEnd();
+				}
+			}, [isDone, typewriter.skipToEnd]);
```
tevanburen: I don't think this effect is necessary because isDone is calculated based on our `activeIndex` in response-message-text, but that itself is controlled base on our onComplete - basically, this component is advancing activeIndex itself, so it doesn't also need to react to changes in activeIndex

packages/playground/src/components/message/response-message-text.tsx
```diff
@@ -1,303 +1,102 @@
...
+
+					return (
+						<ResponseMessageTextMd
+							key={chunk.key}
+							content={chunk.content}
+							isFinalized={chunk.isFinalized}
+							isActive={isActive}
+							isDone={isDone}
+							message={message}
+							isLast={isLast && idx === chunks.length - 1}
```
tevanburen: i'm not convinced this calculation is correct, because chunks could still be growing right? at a glance maybe it just needs `&& chunk.isFinalized`

packages/playground/src/components/message/response-message-text/response-message-text-md.tsx
```diff
@@ -0,0 +1,201 @@
...
+
+			// Once the entire message has finished streaming, skip the typewriter to
+			// end so the full chunk content is shown immediately rather than slowly
+			// trickling in. We intentionally wait for the whole message (not just
+			// this chunk) to be done — intermediate chunks should animate naturally
+			// at their own pace while the LLM is still producing tokens.
+			const isStreaming = isLast && message.isThinking;
+			// biome-ignore lint/correctness/useExhaustiveDependencies: intentionally fires only when isStreaming flips — adding isActive/newContent.length/typewriter.skipToEnd would cause skipToEnd to fire on every token, breaking the typewriter
```
tevanburen: I think this comment is outdated at best

packages/playground/src/components/message/response-message-text/response-message-text-md.tsx
```diff
@@ -0,0 +1,201 @@
...
+			useEffect(() => {
+				if (!isActive) return;
+				if (isStreaming) return;
+				if (!isFinalized) return;
+				if (newContent.length === 0) return;
+				// Skip to end — this sets renderedLength = newContent.length and
+				// stops isRunning, which will trigger the caught-up effect below.
+				typewriter.skipToEnd();
+			}, [isStreaming, isFinalized]);
+
```
tevanburen: IMPORTANT

I think many of these useEffects are doing similar things and stepping on each other's toes. We should dial in the difference between `typewriter.skipToEnd` vs `onComplete` - it's actually possible that skipToEnd is no longer necessary

tevanburen:
Most important comment was the one I marked as IMPORTANT

packages/playground/src/hooks/use-markdown-typewriter.ts
kellogg9: It's not necessary for HTML, however, all other code fences (python, ts, ```sql, etc.) still live inside md chunks and get animated by the typewriter.

packages/playground/src/hooks/use-markdown-typewriter.ts
kellogg9: It came to me in a dream

packages/playground/src/components/message/response-message-text.tsx
tevanburen, line 1: Clean

packages/playground/src/components/message/response-message-text/response-message-text-md.tsx
```diff
@@ -0,0 +1,139 @@
...
+	status: ChunkStatus;
+
+	/** The response message (for room access and isThinking). */
+	message: ResponseMessageStore;
+
+	/**
+	 * True when this is the last part of the last message.
+	 * Drives isPreviewLoading for code blocks (Mermaid, CodePreviewBlock).
+	 */
+	isLast: boolean;
```
tevanburen: I'm not convinced that we need this, take a look at what it actually does

packages/playground/src/components/message/response-message-text/response-message-text-md.tsx
```diff
@@ -0,0 +1,139 @@
...
+				typewriter.rendered.length < newContent.length)
+				? contentOnMountRef.current + typewriter.rendered
+				: content;
+
+		// ── Markdown components ───────────────────────────────────────────────────
+		const isPreviewLoading = isLast && message.isThinking;
+		const components = useMemo(
+			() => createMarkdownComponents(message.room, isPreviewLoading),
+			[message.room, isPreviewLoading],
+		);
```
tevanburen, line 66: This may not need to be memoized

packages/playground/src/components/message/response-message-text/response-message-text-md.tsx
```diff
@@ -0,0 +1,139 @@
...
+				typewriter.start();
+			}
+		}, [
+			status,
+			typewriter.isTyping,
+			typewriter.rendered.length,
+			newContent.length,
+			onComplete,
+			typewriter.start,
+		]);
```
tevanburen, line 112: This useEffect's logic is clean, in fact it's kinda so clean that it may not even need to be a useEffect. The fact that it's depending on typewriter.rendered.length for example, well that's gonna be changing almost every re-render anyway. We could instead just have it as a function, even an IIFE like 
```
(() => {
  if (status !== "active") return;
  // etc
})();
```

This proposed change would also mean that onComplete need not be stable, which would allow us to clean up that callback / memoization in response-message-text if we wanted

packages/playground/src/hooks/use-markdown-typewriter.ts
tevanburen: changes to `contentRef.current` do not force re-renders of this component. It is theoretically possible for isTyping then to be stale. What is the benefit of depending on contentRef.current for this rather than content? same question goes for the change in the useEffect

packages/playground/src/components/room/room-content.tsx
tevanburen, line 1: Have not scrutinized this in depth but the comments look logical

packages/playground/src/components/message/response-message-text/response-message-text-html.tsx
tevanburen, line 1: Have not looked at this file yet, will iterate on response-message-text-md first, bc I assume the solution here will be very very similar to there

kzsb03 (approved):

PR #3304, opened 2026-06-09
feat(client): added all tests for query import cell

boredcod:
## Description
Added a test for query import cell

## How to Test

1. cd libs/renderer   
2. pnpm test src/testing/cell-defaults/query-import-cell.spec.tsx

all tests should pass


memisrose (approved):
Image

PR #3342, opened 2026-06-23
Feature gate

travon-speller:
## Description

implements the reactors here to be able to create/edit feature gates https://github.com/SEMOSS/Semoss/pull/2659

## Changes Made

new page and route for feature gates

## How to Test

1. Route to the features tab
2. Select the app you want to feature gate and create a new flag
3. Create/Add versions and add users 
4. then test the flag in the FE UI for that app or by using the reactor

## Notes

relies on https://github.com/SEMOSS/Semoss/pull/2659


johbaxter:
Is this stale/no longer an ask? @travon-speller

PR #3369, opened 2026-07-08
File Upload Error 1570

SunTheDev:
Description

  When a user drag-and-drops a Word document (or any file) that is currently open in another program, the upload
  silently fails with a net::ERR_FAILED network error — no feedback is shown to the user. This PR adds a clear error  toast to match the existing behavior of the file picker path, which already shows a "This file is in use" message in that scenario.
  
  Changes Made

  - packages/playground/src/stores/room/room.store.ts — In the askMessage upload catch block, detect network-level  
  upload failures (TypeError: Failed to fetch / net::ERR_FAILED) and show a toast: "This file is in use. Close the  
  file in the other program and try again." The toast is raised from the store itself so it fires regardless of     
  whether askMessage is awaited or called fire-and-forgot (as it is in the new chat flow).

  How to Test

  1. Open a .docx file in Microsoft Word (keep it open)
  2. In the chat, drag and drop that same file into the "Drop files here to attach to message" drop zone
  3. Type any message and hit Send
  4. Expected: A toast error appears saying "Uploaded file(s) currently in use by another program. Close the program and re-upload the file(s) in a new chat."
  5. Expected (before fix): The baseUpload network call fails silently with net::ERR_FAILED, no error is shown, and 
  the file is not recognized by the chat
  6. Close the file in Word, repeat steps 2–3
  7. Expected: File uploads and is recognized successfully

  Notes

  - The fix lives entirely in the store so it covers both the new chat flow (new-room-page.tsx calls askMessage     
  fire-and-forgot with no .catch()) and the existing room flow (room-content.tsx awaits it)
  - The file picker path (Windows Explorer dialog) was already blocking locked files at the OS level before the     
  browser received them — this fix brings drag-and-drop to parity with that behavior
  - Detection targets TypeError (the JS error class for network failures) and message strings Failed to fetch /     
  ERR_FAILED / NetworkError to cover Chrome, Edge, and Firefox variants


packages/playground/src/stores/room/room.store.ts
```diff
@@ -986,6 +980,21 @@ export class RoomStore {
...
 				const uploaded = response.data;
 
+				// If files were sent but the server returned nothing, the files
+				// couldn't be read — most likely locked by another program (e.g.
+				// a .docx open in Word). Surface this as an UploadError so the
+				// caller can show a "file is in use" message instead of silently
+				// proceeding with no attachment.
+				if (uploaded.length === 0) {
+					const uploadError = new Error(
+						"Uploaded file(s) currently in use by another program. Close the program and re-upload the file(s) in a new chat.",
```
kzsb03: can we move to support internationalization

packages/playground/src/stores/room/room.store.ts
```diff
@@ -1029,6 +1038,29 @@ export class RoomStore {
...
+			const errMsg = (e as Error)?.message ?? "";
+			const isNetworkOrUploadFailure =
+				errMsg.includes("Failed to fetch") ||
+				errMsg.includes("ERR_FAILED") ||
+				errMsg.includes("File is in use") ||
+				errMsg.includes("NetworkError");
+
+			if (isNetworkOrUploadFailure || e instanceof TypeError) {
+				toast.error(
+					"Uploaded file(s) currently in use by another program. Close the program and re-upload the file(s) in a new chat.",
```
kzsb03: same as above

kzsb03 (changes requested):

packages/playground/src/stores/room/room.store.ts
```diff
@@ -1033,6 +1050,27 @@ export class RoomStore {
...
+
+			// Show the "file is in use" toast here — in the store — so it fires
+			// regardless of whether askMessage was awaited or fire-and-forgot
+			// (new-room flow calls it without await and swallows the thrown error).
+			const errMsg = (e as Error)?.message ?? "";
+			const isNetworkOrUploadFailure =
+				errMsg.includes("Failed to fetch") ||
+				errMsg.includes("ERR_FAILED") ||
+				errMsg.includes("File is in use") ||
+				errMsg.includes("NetworkError");
```
tevanburen, line 1063: These won't actually catch the errors we're throwing

packages/playground/src/stores/room/room.store.ts
```diff
@@ -8,6 +9,7 @@ import {
 	uploadInsight,
 } from "@semoss/sdk/react";
 import { FlexLayout, type ThemeMap } from "@semoss/shared";
+import { toast } from "@semoss/ui/next";
```
tevanburen, line 12: i don't think we should be using i18n or toast inside the store, i'd prefer to do it in the caller instead

packages/playground/src/stores/room/room.store.ts
```diff
@@ -1033,6 +1050,27 @@ export class RoomStore {
...
+				errMsg.includes("ERR_FAILED") ||
+				errMsg.includes("File is in use") ||
+				errMsg.includes("NetworkError");
+
+			if (isNetworkOrUploadFailure || e instanceof TypeError) {
+				toast.error(getI18n().t("room:errors.fileInUse"));
+				// Tag and re-throw so awaited callers can also react if needed
+				const uploadError = new Error((e as Error).message);
+				uploadError.name = "UploadError";
+				(uploadError as Error & { fileNames: string[] }).fileNames =
```
tevanburen, line 1070: no code reads this value

ekim1707:
@tevanburen @kzsb03 i pushed some changes. could i request another review please?

kzsb03 (approved):

PR #3378, opened 2026-07-09
fix: make default model reflect immediately on new room page upon switch

ekim1707:
## Description
This is a fix on earlier 1357 default model branch to make the new room page default model dropdown reflect immediately as soon as user changes default model on the Models page

## Changes Made
refreshProfileDefaultModel method added to chat.store.ts that makes backend GetUserInfo() call. Then useEffect added to trigger that method upon mount on new-room-page so that the default model change reflects instantly

## How to Test
1. Change default model in Models page
2. Navigate to the new room page and see if the default model change was reflected

## Notes
If the default model is changed on the UI, the change is not reflected in the Settings > My Profile page until after logout/login. This looks to be due to MobX in-memory state that only provides default model once upon login and is never resynced.


packages/playground/src/stores/chat/chat.store.ts
```diff
@@ -478,6 +495,51 @@ export class ChatStore {
...
+	/**
+	 * Re-fetch the user's profile default text-generation model from the backend
+	 * and update the selected model if it has changed. Called when the new-room
+	 * page mounts so any change made in the token-usage page (embed) is picked
+	 * up without requiring a logout/login cycle.
+	 */
+	refreshProfileDefaultModel = async (): Promise<void> => {
+		try {
+			// Use GetUserMetadata to bypass the GetUserInfo cache
+			const result = await this._actions.run<[Record<string, string[]>]>(
```
tevanburen, line 507: should type this better, rather than leaving it generic Record

packages/playground/src/stores/chat/chat.store.ts
```diff
@@ -210,44 +214,57 @@ export class ChatStore {
...
-				this._store.user = {
-					id: providerData.id,
-					name: providerData.name,
-					lastLogin: providerData.lastLogin,
-				};
-			});
+			// Run GetUserInfo and GetUserMetadata in parallel.
+			// GetUserMetadata bypasses the cache that GetUserInfo has, so it
+			// always returns the latest saved text-generation-model value.
+			const [userInfoResult, userMetaResult] = await Promise.allSettled([
```
tevanburen, line 220: rather than like this, should just send both pixel calls in the same request like `run(META | GetUserInfo(); META | GetUserMetadata(); )`

packages/playground/src/stores/chat/chat.store.ts
```diff
@@ -616,6 +678,10 @@ export class ChatStore {
 
 		runInAction(() => {
 			const { output } = pixelReturn[0];
+
+			// Cache the full list so setProfileDefaultModel() can resolve an id to an Engine later
```
tevanburen, line 682: this is either a typo or no longer used

packages/playground/src/stores/chat/chat.store.ts
ekim1707: i'm not that familiar with typing, but does this look ok?

  const result = await this._actions.run<[{ "text-generation-model"?: string | string[] }]>(
      `META | GetUserMetadata();`,
  );

packages/playground/src/stores/chat/chat.store.ts
ekim1707: getUser = async (): Promise<void> => {
      try {
          const result = await this._actions.run<
              [
                  Record<string, { id: string; name: string; lastLogin?: string }>,
                  UserMetadata,
              ]
          >**(`META | GetUserInfo(); META | GetUserMetadata();`);**

put the both pixels calls in one request in bold. does this check out?

packages/playground/src/stores/chat/chat.store.ts
```diff
@@ -478,6 +481,51 @@ export class ChatStore {
...
+	refreshProfileDefaultModel = async (): Promise<void> => {
+		try {
+			// Use GetUserMetadata to bypass the GetUserInfo cache
+			const result = await this._actions.run<
+				[{ "text-generation-model"?: string | string[] }]
+			>(`META | GetUserMetadata();`);
+
+			const meta = result.pixelReturn[0].output;
+			const metaValue = meta?.["text-generation-model"];
+			const newModelId = Array.isArray(metaValue)
```
tevanburen, line 499: you could make this a helper (cuz it's used in 2 places) but not a huge deal

packages/playground/src/stores/chat/chat.store.ts
tevanburen: should be refreshProfileDefaultModel I think

kzsb03 (approved):

PR #3391, opened 2026-07-13
fix: fixed the don't support landscape screen to not show on desktop …

ekim1707:
…mode

## Description
Prevents mobile landscape screen from showing on desktop

## Changes Made
Made changes on the landscape-restriction.tsx page that allowed toggle between mobile landscape mode and normal.

## How to Test
1. From any page in browser, make the width as small as possible. The landscape page should not show.
2. Go to Developer Tools and use the phone toggle device toolbar. If you put some devices in landscape mode, the landscape page should show.

## Notes
Screenshot 2026-07-13 174907
Screenshot 2026-07-13 174949


libs/i18n/src/resources/locales/ar/playground/mobile.json
tevanburen, line 1: should give these real translations

packages/playground/src/components/common/landscape-restriction.tsx
```diff
@@ -35,64 +58,31 @@ export const LandscapeRestriction = () => {
...
-				backgroundColor: "white",
-				zIndex: 9999,
-				display: "flex",
-				flexDirection: "column",
-				alignItems: "center",
-				justifyContent: "center",
-				padding: "2rem",
-				textAlign: "center",
-			}}
+			data-testid="landscape-restriction-overlay"
```
tevanburen: should omit these if not actually used

packages/playground/src/components/common/landscape-restriction.tsx
```diff
@@ -35,64 +58,31 @@ export const LandscapeRestriction = () => {
...
+			className={cn(
+				"fixed inset-0 z-[9999]",
+				"flex flex-col items-center justify-center",
+				"bg-white px-8 text-center",
```
tevanburen: make sure to use theme tokens like bg-background, hardcoding the colors breaks dark mode

tevanburen (commented):
I'm not actually able to get the screen to come up at all now, like I am on dev

ekim1707:
> I'm not actually able to get the screen to come up at all now, like I am on dev

i think that's supposed to be intended behavior, unless you were on the phone emulator. On desktop, isMobileDevice is always false, so isLandscape stays false and the component returns null.

ekim1707:
@tevanburen i updated this branch. please re-review. thank you!

tevanburen:
> > I'm not actually able to get the screen to come up at all now, like I am on dev
> 
> i think that's supposed to be intended behavior, unless you were on the phone emulator. On desktop, isMobileDevice is always false, so isLandscape stays false and the component returns null.

Already merged, but I was using the phone emulator

PR #3439, opened 2026-07-16
Customize compaction behavior

ckelly17:
## Summary

- Adds a **Compaction Options** picker to the token-counter popover in the chat input bar, letting users choose how the conversation is compacted before triggering it manually
- The default strategy is theme-configurable via `defaultCompactionStrategy` (`"SUMMARY"` | `"TOOL_PRUNE"` | `"AUTO"`) and initializes the picker on room load
- Strategies are exposed under a collapsible **Compaction Options** section with labeled radio buttons and info tooltips explaining each approach; the top-level **Compact** button triggers compaction immediately with the selected strategy

## What's included

- `CompactStrategyPicker` inline component in `room-input.tsx` — single Compact button + collapsible strategy radio group
- `compactionStrategy` state in `room-content.tsx` seeded from `root.theme.defaultCompactionStrategy`
- `compactMessages(strategy?)` in `room.store.ts` passes the chosen strategy to the pixel (`compactionTypes=["<strategy>"]`); omits the param when strategy is `AUTO`
- `defaultCompactionStrategy` wired through `ThemeMap` → `root.store.ts` → component state

## Test plan

- Open a room; token counter appears when tokens are in use
- Hover the token pie chart — popover opens with a **Compact** button and collapsed **Compaction Options**
- Expand Compaction Options — three radio buttons (Summarize, Prune Tools, Auto) with tooltip icons
- Select a strategy, click Compact — toast confirms success; conversation is compacted with the chosen method
- Set `defaultCompactionStrategy: "TOOL_PRUNE"` in `theme.local.json`, restart dev server — picker initializes to Prune Tools
- Verify send button and chat input are never disabled or replaced due to token usage

image
image


libs/i18n/src/resources/locales/ar/playground/room.json
tevanburen, line 1: will need to translate these other files too

libs/shared/src/types.ts
tevanburen, line 1: should also add to the theme map editor page in /client

tevanburen:
Waiting on #3528 and #3529 btw

kzsb03 (approved):

PR #3445, opened 2026-07-17
Search Bar Enhancement: wire message content search

Kaveh-Vakili:
Goes with https://github.com/SEMOSS/Semoss/pull/2764

## Description
Formulated a more robust solution for the playground search bar that is able to search for content within different rooms.

## Changes Made
- global-nav.tsx: fire SearchRoomMessages when searching, merge content-match rooms into bucketed list, fix empty-state check to use mergedRooms so content matches are not hidden
- chats-page.tsx: same SearchRoomMessages useEffect pattern, merge content-match-only rooms into visibleRooms

## How to Test
1. Create multiple different rooms with conversations
2. With the Playground All Chats search bar, search for content within those rooms


packages/playground/src/pages/chats-page.tsx
```diff
@@ -84,6 +84,39 @@ export const ChatsPage = observer(() => {
...
+					(r) => r.room_id && seen.size !== seen.add(r.room_id).size,
+				)
+				.map((r) => ({
+					ROOM_ID: r.room_id,
+					ROOM_NAME: r.room_name ?? "",
+					DATE_CREATED: r.date_created ?? "",
+				}));
+		},
+		{ limit: 50 },
+		[debouncedSearch, chat.keys.roomCounter],
```
tevanburen, line 113: we should call refresh()/reset() directly on roomCounter change, like getRooms already does, instead of just adding it to useIteratorPixel's deps array

kunal0137:
Confirmed locally: pagination is applied to matching messages, but the UI displays matching rooms.
For example, assume Room A contains 50 separate messages containing “spain,” while Room B contains one message containing “spain.” With limit=50, the backend can return all 50 rows from Room A. The UI then deduplicates those rows and displays only Room A, so Room B is incorrectly missing.
In my local data, “spain” matched 136 messages across 44 rooms. The first 50 message rows represented only 18 rooms, leaving 26 valid matching rooms out of the results.
The backend should reduce the results to one row per room before applying limit and offset. When testing, ensure the room name does not contain the search term; otherwise the separate room-name search masks this issue.


additionally infinite or additonal scroll doesnt work after.

packages/playground/src/pages/chats-page.tsx
```diff
@@ -77,10 +77,53 @@ export const ChatsPage = observer(() => {
...
+			message_id: string;
+			room_name: string;
+			date_created: string;
+		}[],
+		RoomItem
+	>(
+		(limit, offset) =>
+			debouncedSearch
+				? `META | SearchRoomMessages(search="<encode>${debouncedSearch}</encode>", project="SYSTEM__PLAYGROUND", limit=${limit}, offset=${offset}, includeMessageText=false);`
+				: `META | Return(value=[]);`,
```
tevanburen: This isn't a reactor

packages/playground/src/pages/chats-page.tsx
```diff
@@ -77,10 +77,53 @@ export const ChatsPage = observer(() => {
...
+			room_id: string;
+			message_id: string;
+			room_name: string;
+			date_created: string;
+		}[],
+		RoomItem
+	>(
+		(limit, offset) =>
+			debouncedSearch
+				? `META | SearchRoomMessages(search="<encode>${debouncedSearch}</encode>", project="SYSTEM__PLAYGROUND", limit=${limit}, offset=${offset}, includeMessageText=false);`
```
tevanburen: could we add SYSTEM__PLAYGROUND to constants

packages/playground/src/pages/chats-page.tsx
tevanburen: ^

tevanburen (changes requested):

packages/playground/src/pages/chats-page.tsx
Kaveh-Vakili: Should SearchRoomMessages be called with META | like the other GetPlaygroundRooms calls in this file, or without it? Keeping it consistent either way — happy to revert if META is the right pattern for useIteratorPixel.

tevanburen: Yeah, should be called with META, but fyi `Return` is not a reactor

packages/playground/src/pages/chats-page.tsx
```diff
@@ -77,10 +78,53 @@ export const ChatsPage = observer(() => {
...
+			room_id: string;
+			message_id: string;
+			room_name: string;
+			date_created: string;
+		}[],
+		RoomItem
+	>(
+		(limit, offset) =>
+			debouncedSearch
+				? `META | SearchRoomMessages(search=["<encode>${debouncedSearch}</encode>"], project=["${SYSTEM__PLAYGROUND}"], limit=[${limit}], offset=[${offset}], includeMessageText=[false]);`
```
tevanburen: Can you use JSON.stringify for these params? like `search=${JSON.stringify(debouncedSearch)}`

packages/playground/src/pages/chats-page.tsx
tevanburen: ^^^

tevanburen:
Kaveh are you seeing my comments about `we should call refresh()/reset() directly on roomCounter change, like getRooms already does, instead of just adding it to useIteratorPixel's deps array`

Kaveh-Vakili:
should be good now. just saw this one

tevanburen:
Pushed a commit fixing the chats-page loading flash (spinner/no-chats flashing on every reload). Root cause was in the shared useIteratorPixel hook - it clears data on every dependency change, unlike useIteratorApi which deliberately doesn't. Fixed that plus reworked chats-page's loading state to dim in place instead of clearing.

Made the fix directly rather than flagging it - it's existing dev behavior, not something from this PR.

tevanburen (approved):

tevanburen:
FE looks good to me, BE still has a comment outstanding (BE PR needs to be merged first)

PR #3493, opened 2026-07-27
fix(client): stop gating project creation on the flat admin-only flag

Kaveh-Vakili:
## Description
configStore.isEngineOperationAvailable("PROJECT", "add") only checks a single boolean (adminOnlyProjectAdd) with no awareness of project type. The backend was recently updated ([companion PR in Semoss]) to allow specific project types through even when that flag is enabled, but three places in the UI were still using this flat, non-type-aware check to pre-emptively block or hide creation entirely — making it impossible for non-admins to create Agent/Skill/Blocks apps even when the backend correctly allows them. This PR removes those incorrect gates so the UI stops guessing and lets the backend be the actual source of truth.

## Changes Made
project-catalog.tsx — the "Add App/Skill/Agent" button was hidden entirely whenever adminOnlyProjectAdd was true, regardless of type. Now always shown.
create-app-page.tsx — redirected away from /app/new entirely for the same reason, blocking both Code and Blocks creation. Redirect removed.
landing-page.tsx — redirected the whole home page to itself, rendering nothing, for any non-admin whenever adminOnlyProjectAdd was true. Redirect removed.

## How to Test
Set ADMIN_ONLY_PROJECT_ADD=true on the backend (with ADMIN_ONLY_PROJECT_ADD_EXEMPT_TYPES including WORKSPACE,SKILL,BLOCKS), log in as a non-admin user.
Confirm "Add Agent"/"Add Skill"/"Add App" buttons are visible, and creating an Agent, Skill, or Blocks app succeeds.
Confirm attempting to create a CODE app still correctly fails with the backend's "limited to only admins" error — this UI change doesn't bypass the actual restriction, it just stops incorrectly hiding options the backend would allow.

## Notes
Companion to the Semoss PR adding ADMIN_ONLY_PROJECT_ADD_EXEMPT_TYPES.
Depends on that backend PR to be meaningful — without it, adminOnlyProjectAdd=true still blocks everything at the backend level regardless of these UI fixes.

https://github.com/SEMOSS/Semoss/pull/2807


Kaveh-Vakili:
Open question for reviewers

The button in project-catalog.tsx is currently shown to all users regardless of adminOnlyProjectAdd. Options for how to handle this:

Hide entirely — isEngineOperationAvailable(...) ? <Button> : null
Disable + tooltip — button always visible, disabled with "Only admins can add projects" tooltip when flag is on
Show always, block at the page — button navigates freely, create-app-page.tsx shows a permission banner for non-admins (also handles direct URL access)
Show always, no gating — current state in this PR

ckelly17:
@Kaveh-Vakili can you link backend PR?

Kaveh-Vakili:
> @Kaveh-Vakili can you link backend PR?

https://github.com/SEMOSS/Semoss/pull/2807

ckelly17:
Put it at the top of this PR

PR #3495, opened 2026-07-27
feat: allow iconDark, appDark, and logoDark through theme

ekim1707:
## Description
Allows coordinated elsa-beta dark mode with frontend.

## Changes Made
Dark asset support for playground regarding the outer shell's sidenav and app logo:
 - New optional theme fields (iconDark, appDark, logoDark) allow dark-mode-specific image URLs to be configured per nav item and logo.
 - New useIsDark hook detects dark mode by watching the dark class on <html> via MutationObserver.
 - AppLogo and GlobalNavItem now use the dark asset when provided, and only fall back to the CSS inversion filter when no dark asset exists.
 - GlobalNav threads iconDark down to each nav item, and RootStore initializes the new fields.

## How to Test
iconDark, appDark, and logoDark will override inversions.
Toggle dark mode without iconDark and make sure icons invert. Add an iconDark, and make sure the inversion get overwritten.
Do the same for appDark


packages/playground/src/components/common/app-logo.tsx
```diff
@@ -14,24 +14,22 @@ interface AppLogoProps {
...
-					className="dark:brightness-0 dark:invert"
-				/>
-			) : (
-				<img
-					alt={t("images.logoAlt")}
-					src={root.theme.images.logo || logoImage}
-					className="dark:brightness-0 dark:invert"
-				/>
-			)}
+			<img alt={t("images.logoAlt")} src={src} className={imgClass} />
```
kzsb03, line 34: I think we need to preserve the 'full' prop passed as they render different things. Theres the full svg of the app name and then the condensed icon

kzsb03 (changes requested):

packages/playground/src/components/common/app-logo.tsx
ekim1707: doesn't the 'full' prop get preserved lines 17-26?

kzsb03: confirmed. can we default to light if we don't have images.appdark and images.logodark?

ekim1707: Hm. i thought it was already defaulted to light mode, but how about this? i might just be missing the picture here	

	const isDark = useIsDark();

	const lightSrc = full
		? root.theme.images.app || appImage
		: root.theme.images.logo || logoImage;
	const darkSrc = full
		? root.theme.images.appDark
		: root.theme.images.logoDark;
	const src = isDark && darkSrc ? darkSrc : lightSrc;
	const imgClass = "";

tevanburen:
Do we need useIsDark here, or could AppLogo and GlobalNavItem just use useTheme() from @semoss/ui/next instead, the same way login-page.tsx, error-page.tsx, workspace-page.tsx, and the others already do? Both components sit inside the <ThemeProvider> tree in app.tsx, so resolvedTheme === "dark" would give the same answer without a second, DOM-watching mechanism living alongside the one the rest of the app already relies on.

ekim1707:
> Do we need useIsDark here, or could AppLogo and GlobalNavItem just use useTheme() from @semoss/ui/next instead, the same way login-page.tsx, error-page.tsx, workspace-page.tsx, and the others already do? Both components sit inside the tree in app.tsx, so resolvedTheme === "dark" would give the same answer without a second, DOM-watching mechanism living alongside the one the rest of the app already relies on.

I see what you mean. i'll try to push some changes soon

packages/playground/src/components/common/app-logo.tsx
tevanburen: I think Kyle means that if there is no appDark/logoDark, we should just use app/logo instead (even if mode === "dark") (that way if an env doesnt have appDark/logoDark set, they still get some logos to appear)

packages/playground/src/hooks/index.ts
```diff
@@ -2,6 +2,7 @@ export * from "./use-active-index";
 export * from "./use-chat";
 export * from "./use-global-breadcrumbs";
 export * from "./use-graceful-errors";
+export * from "./use-is-dark";
```
tevanburen: leftover, need to remove

packages/playground/src/components/common/app-logo.tsx
ekim1707: but if there's no appDark/logoDark, then in the above code, darkSrc is undefined, and const src resolves to lightSrc, so wouldn't that default it to light mode?

tevanburen: ah ok valid

ekim1707:
> > Do we need useIsDark here, or could AppLogo and GlobalNavItem just use useTheme() from @semoss/ui/next instead, the same way login-page.tsx, error-page.tsx, workspace-page.tsx, and the others already do? Both components sit inside the tree in app.tsx, so resolvedTheme === "dark" would give the same answer without a second, DOM-watching mechanism living alongside the one the rest of the app already relies on.
> 
> I see what you mean. i'll try to push some changes soon

pushed these changes as well @tevanburen if you'd like to confirm

https://github.com/SEMOSS/semoss-ui/pull/3495/commits/e67ee9d89d49d6c96724b490b8d9c9d7c5eeb5c2

kzsb03 (approved):

PR #3514, opened 2026-08-03
fix: correct projectid reference for mcp usage tab

themaherkhalil:


themaherkhalil:
Fixed with https://github.com/SEMOSS/semoss-ui/commit/f1b1e5bd8022a263d687ab4dc70a548a1e808f4b

PR #3515, opened 2026-08-04
Fix add member modal

tevanburen:
## Description
Fixes several Add Members modal UX bugs and consolidates all permission-related APIs and types (`Role`, `User`, `PostUser`, `UserAccessRequest`) into `@semoss/sdk` as the single source, since they were duplicated across `libs/shared` and `packages/client` with drifting shapes and argument orders.

## Changes Made
- **Add Members modal:** debounce search, stop layout jumping, fix an empty-list race on open, add i18n
- **`useIteratorApi`:** relocated to `@semoss/sdk/react` so it's shared instead of copied per package
- **Permissions API:** all project/engine permission CRUD now lives only in `libs/sdk/src/api/permissions.ts`; standardized engine argument order to match project; fixed `UserAccessRequest`'s shape to match what the backend actually reads (`requestid`/`userid`, not `id`)
- **Type dedup:** `Role`/`User`/`PostUser`/`UserAccessRequest` existed in three places — now only in sdk, everything else imports from there
- **Bug fix:** adding a member no longer triggers a full project refresh that blanked the page and reset the active tab; editing your own permission still does, since that's the case that actually matters

## How to Test
1. Add Members modal — search should debounce with no flicker, and Add should succeed without the page flashing or resetting tabs.
2. Engine Members tab — edit, search, and remove a member; search previously didn't actually filter (wrong query param), now it does.

## Notes
The type duplication (`Role`/`User`/`PostUser`/`UserAccessRequest`) was flagged mid-review and fixed here rather than deferred.


libs/sdk/src/js-frameworks/react/hooks/use-iterator-api.ts
neelneelneel, line 1: This should go into sdk. UI should be dumb ui components only

libs/shared/src/components/members/model-restriction-fields.tsx
```diff
@@ -0,0 +1,160 @@
...
+	Input,
+	Label,
+	Select,
+	SelectContent,
+	SelectItem,
+	SelectTrigger,
+	SelectValue,
+} from "@semoss/ui/next";
+
+const formatNum = (val: string): string => {
```
neelneelneel: Can these go into the utility folder? We can reuse across the platform

libs/shared/src/components/members/add-members.tsx
```diff
@@ -80,86 +77,93 @@ export const AddMembersOverlay = ({
...
-			const perm = (response?.data as { permission?: string })
-				?.permission;
-			if (perm) setUserPermission(perm);
-		}
+		const fetchMyPermission = async () => {
+			try {
+				const endpoint = isProject
+					? `project/getUserProjectPermission?projectId=${id}`
+					: `engine/getUserEnginePermission?engineId=${id}`;
+				const response = await apiGet(
```
neelneelneel: can we pull from /api - there are two in sdk + shared. I'll ping you.

neelneelneel (changes requested):

PR #3518, opened 2026-08-04
feat: added temperature to theme db

ekim1707:
## Description
This PR allows for the "Temperature" slider in the Room Settings to be displayed or not displayed according to the enableTemperature feature flag from theme db

## Changes Made
Much of the changes were undoing previous changes that took Temperature out of the code, and also incorporating the feature flag from theme db

Here's a summary of all the changes:

  UI (room-options-form.tsx)

  - Added a temperature Slider (0–1, step 0.01) at the bottom of Room Settings
  - Gated behind root.theme.featureFlags?.enableTemperature so it only shows when the flag is on

  Type definitions (libs/shared/src/types.ts)

  - Added enableTemperature?: boolean to the featureFlags type
  - Added temperature?: number to defaultRoomSettings type

  Translations (libs/i18n/.../playground/room.json)

  - Added "temperatureLabel" key to all 7 languages

  Root store (root.store.ts)

  - Added enableTemperature: false as the default so it's always defined even if the backend doesn't send it

  Room store (room.store.ts)

  - Added temperature?: number field to the room options interface
  - Initialized it as undefined by default

  API call (response-message.store.ts)

  - The AskPlayground pixel call now includes temperature in paramValues when the flag is on and a value is
  set — so the model actually receives it

  New room initialization (new-room-page.tsx)

  - Seeds temperature from defaultRoomSettings.temperature in the theme when creating a new room, if the flag
  is on

## How to Test
1. Load semoss-ui and toggle enableTemperature true and false from either theme.local.json or theme db. Make sure the temperature slider displays appropriately for each toggle.
2. Send a prompt to Elsa and check the AskPlayground reactor in the network tab. It should have a "temperature" parameter equal to the value of the slider on the UI

Screenshot 2026-08-04 144847
Screenshot 2026-08-04 145023


ckelly17:
@ekim1707 can you check on this?

Actually, seems like all build checks are failing. Probably fine

PR #3527, opened 2026-08-06
Sdk playground api

johbaxter:
GitHub Copilot only scans for skills inside your workspace's skills (or .agents/skills/). It does not crawl node_modules.

So the flow is:

The extraction step is always required. That's why the README documents it — consuming teams need to run it once (or wire it into postinstall to make it automatic on every install/upgrade).

The shortest path for a team to get zero-friction setup is the postinstall approach from the README:

---

Yes — both paths are fully independent and either can be used without touching the other.

Raw API path — import individual functions:

image

Room construct path — import the class/factory:

image

Both are exported from the same @semoss/sdk entry point, so there's no separate install or config. A developer can even mix them — e.g. use [createRoom] for conversation management but drop down to [askPlayground]( directly for a one-off custom call where they need full control.

The [SKILL.md] already has the comparison table calling this out explicitly in the "Room construct (managed state)" section. If you want, I can make that choice point more prominent — e.g. move the comparison table to the very top of the skill so it's the first thing a developer reads.


johbaxter:
image

Latest commit

PR #3534, opened 2026-08-11
Agent streaming

### LINKED ISSUE #3535: Agent harness SDK updates

tevanburen:
## Description
Adds a full agent mode (RunAgent) to Playground — a server-driven agent harness loop that streams messages, tool calls, and subagent activity back to the client, as an alternative to the existing client-orchestrated chat flow. Requires the companion Semoss backend `agent-streaming` branch to function.

## Changes Made

**SDK (`libs/sdk`)**
- New `agent.ts` / `agent.types.ts` / `agent-subscription.ts`: pure API wrappers for `RunAgent`, `GetAgentRun`, `GetSubagentRuns`, and tool-decision pixels, plus a `subscribeAgentRun` polling/reconciliation subscription and shared item-event merging — decoupling the wire protocol from Playground-specific state.

**Agent mode entry & UI (`room-input.tsx`, `new-room-page.tsx`, `constants.ts`, slash-command context)**
- `/agent-harness` (and hidden `/harness`) slash commands to switch a room into agent mode.
- Consolidated room-input mode/workspace/exit chips into one, added an agent-mode indicator, and synced agent mode into the new-room store so a mode set before persistence isn't lost.
- Decoupled workspace selection from chat vs. agent mode.

**Streaming & tool execution (`agent-harness.ts`, `response-message.store.ts`, `response-message-tool*.tsx`, `tools-view.tsx`)**
- Core `agent-harness.ts` store: submits agent runs, applies streamed item events (message/reasoning/tool/subagent) onto the room's message tree, reconnects an in-progress run after a page reload, and resumes a paused run on a tool decision.
- Proper "ask" UI and tool grouping for agent-run tool calls, plus title/open-tool fixes.
- Fixed a stuck-loading bug in `ToolsView` for tools with no resolvable MCP app behind them.

**Subagents**
- Subagent status renders as its own card (styled like the "large" ask-mode tool card) instead of a raw tool pill, with per-status treatment (running / waiting-for-input / completed / failed / cancelled).
- Reconstructs every turn's subagents from the durable `AGENT_RUN` table on reload — not just the tail turn's — positioned under the message that spawned them.
- Clicking a subagent card opens a side panel (reusing the existing sidebar tab system) showing its response/error, kept live via a room-wide lookup.
- Subagent tool calls (spawn/named/check/wait) now render normally, both live and after reload, matching regular tools.

**Cancellation**
- Stop button shows a plain spinner instead of an actionable-looking Stop control during agent-mode turns, since agent runs can't yet be reliably interrupted server-side.

## How to Test
1. In a room, run `/agent-harness` to switch into agent mode.
2. Send a message that triggers tool calls — verify tool calls group/ungroup correctly and ask-mode tools show the approve/decline UI.
3. Ask the model to spawn a subagent (anonymous or named) — a styled subagent card should appear live; clicking it opens a side panel with its response.
4. Refresh mid-run and after a completed run with subagents from an earlier turn — the run should reconnect (or show its correct terminal state), and all historical subagents should reappear under their spawning message.
5. Confirm the Stop button shows a spinner, not a clickable Stop, during an agent-mode turn.

## Notes
- Requires the companion Semoss backend `agent-streaming` branch — subagent `_meta`/execution-mode fixes, the playground room-list filter excluding subagent rooms, and the `GetSubagentRuns`/`StopAgentRun` reactors all live there.
- Agent-run cancellation isn't wired up yet: the backend's cooperative interrupt can be silently swallowed by a blocking socket wait in `SocketClient.executeCommand`, so a working Stop button is follow-up work, tracked separately.
- Skill lives at https://github.com/SEMOSS/Semoss/blob/agent-streaming/project/platform__agent-run/app_root/version/assets/public/SKILL.md


libs/sdk/src/api/agent.ts
```diff
@@ -0,0 +1,389 @@
...
+
+export type * from "./agent.types";
+
+/**
+ * Submit a durable agent run without waiting for it to finish (RunAgent with
+ * wait=false). Poll progress with pollAgentRun(handle.runId) or subscribeAgentRun.
+ */
+export const submitAgentRun = async (
+	params: {
+		roomId: string;
```
neelneelneel, line 35: Can we comment each of the parameters / the return?

libs/sdk/src/api/agent.ts
```diff
@@ -0,0 +1,389 @@
...
+/**
+ * Poll an agent run to completion, delivering item events and durable
+ * snapshots as they arrive. Owns dedup, ordering, reconciliation timing, and
+ * retry/backoff — a transport failure never concludes the run failed.
+ */
+export const subscribeAgentRun = (
+	runId: string,
+	handlers: AgentRunSubscriptionHandlers,
+	options: AgentRunSubscriptionOptions = {},
+): AgentRunSubscription => {
```
neelneelneel: if types aren't used in other places, can we keep it inline?

libs/sdk/src/api/agent.ts
```diff
@@ -0,0 +1,389 @@
...
+		if (stopped) {
+			return;
+		}
+		stopped = true;
+		signal?.removeEventListener("abort", stop);
+		wake?.();
+		// Guard against deleting a newer subscription for the same runId that
+		// superseded this one (e.g. reconnecting after this one already
+		// finished) — only remove the registry entry if it's still this one.
+		if (activeSubscriptions.get(runId) === subscription) {
```
neelneelneel: You shouldn't mix logic with the api - keep the API pure

neelneelneel (changes requested):

johbaxter:
Yeah skill file for this @tevanburen

PR #3585, opened 2026-08-25
feat: adding design.md

neelneelneel:
## Description
Adding Design.md


johbaxter (approved):
Makes sense, more data for agents.md to follow design standards

reysua (approved):
Looks good to me. This addresses the problem where a developer has to manually call the skill.

PR #3588, opened 2026-08-25
Tvb tabbed tool card

### LINKED ISSUE #3587: Formatting updates to default tool view
"can we have a max for the description up top as well and allow the right hand side to actually scroll down for when we have a lot of params and want to view it on the UI"

"what do you think about switching this to tabbed instead of vertical form?"

tevanburen:
## Description
The default tool card mixed the parameters form, raw JSON output, and description into one long scroll, so a tool with a few long fields or a big output was a pain to read. Split it into Description / Inputs / Output tabs, long fields collapse individually, panels scroll with the app's own scrollbar instead of a native one

## Changes Made
- Tabbed tool card (Description/Inputs/Output), shared between MCP tools and server tools like `web_search`
- Multiline strings, arrays, and objects in Inputs can now be collapsed one at a time
- Tabs use the app's ScrollArea instead of native scroll; fixed a bug where a long description got clipped instead of scrolling
- Server tool output now shows error styling on failure - it only handled success before
- Cleared out 4 dead i18n keys

## How to Test
- Make sure the default UI looks decent for tools


johbaxter (approved):
Looks fine to me, just updating the form view for tool invocation.

Next time if you can drop screenshots

PR #3602, opened 2026-08-28
Custom dark mode color

ckelly17:
## Summary
- Adds `darkModeBaseColor` to the theme, allowing the entire dark mode surface palette to be configured from the theme database via a single value. This allows for an implementation to set a particular palette for dark mode without impacting other deployments.
- Introduces `--dark-base` CSS variable in `globals.css`; all dark surface tokens (`--background`, `--card`, `--sidebar`, `--popover`, `--muted`, `--accent`, etc.) are expressed as `calc()` offsets from it — offsets live in CSS, not JavaScript
- When set, the theme applies `--dark-base` as a single inline style on the document root; when unset, it is removed and the browser falls back to the default palette unchanged
- Adds an auto-inversion of logo if no dark mode icon is detected.


ekim1707 (approved):

ekim1707:
tested and approved

kzsb03 (approved):

PR #3614, opened 2026-08-31
refactor(client): drop rhf and zod from package.json

johbaxter:
## Description

- Removed the generic data-driven Function import form (function-import.tsx, function-import-form.tsx, function-import.constants.ts)
- Added shared helpers: name-uniqueness check + createFunctionEngine() pixel/upload helper, plus reusable parameter-list/string-list array fields
- Added 13 dedicated connector form components (AWS, Azure, Brave/Bing Search, Google, Local Python, REST, ServiceNow, etc.)
- Added a connector registry (slug → metadata + Component)
- Added a list/search page and a dynamic :connector detail page
- Wired routing: new → list, new/:connector → connector form

---

TODO:

create section component to preserve styles of form
 
create skill/agent so we only have to run this per each flow once.

## How to Test

2. Expected outcomes


johbaxter:
image

libs/ui/src/next/form.tsx
```diff
@@ -486,6 +488,57 @@ function FormFileDropzone({
...
+	testIdPrefix: string;
+	children: React.ReactNode;
+};
+
+/**
+ * A labeled group of fields: a title + description column beside the field
+ * column, closed off with a separator. Use to group a long form's fields
+ * (e.g. General / Credentials / Settings) instead of one flat field list.
+ */
+function FormSection({
```
johbaxter: This could go into its own file, or we can keep it here for ease with agent

libs/ui/src/next/form.tsx
```diff
@@ -486,6 +488,57 @@ function FormFileDropzone({
...
+	title,
+	description,
+	testIdPrefix,
```
neelneelneel: can we remove this testId? It should be on the input element

packages/client/src/components/import/function/shared/function-engine.utils.ts
```diff
@@ -0,0 +1,103 @@
+import { runPixel } from "@semoss/sdk/react";
+import { z } from "@semoss/ui/next";
+import { uploadFile } from "@/api";
+
+/** Catalog names may only contain alphanumerics, dashes, and spaces. */
+export const CATALOG_NAME_PATTERN = /^[\w\-\s]+$/;
+
+/** Shared zod schema for every connector's "Catalog Name" field. */
+export const catalogNameSchema = z
```
neelneelneel: Can this be moved into a catalog.utility.ts? Both projects / enignes will require this

packages/client/src/components/import/function/shared/function-engine.utils.ts
```diff
@@ -0,0 +1,103 @@
...
+	}),
+);
+
+/** Reusable zod fragment for the FUNCTION_REQUIRED_PARAMETERS metadata field. */
+export const stringListSchema = z.array(z.string());
+
+/**
+ * Check whether a catalog name is already taken (CheckEngineName pixel).
+ */
+export const isEngineNameTaken = async (name: string): Promise<boolean> => {
```
neelneelneel: I think this should be a shared component

neelneelneel: a special input type

johbaxter:
Summary
Redid the field-audit fixes (they'd been reverted) and built out the full 19-connector set against the JSON you pasted as ground truth:

Reapplied to the 5 existing connectors with metadata fields (Brave, Bing, REST, ServiceNow, Local Python): split FUNCTION_NAME/FUNCTION_DESCRIPTION/FUNCTION_PARAMETERS/FUNCTION_REQUIRED_PARAMETERS into their own "Function Metadata" section, dropped the (metadata) label suffix, restored the shared FUNCTION_METADATA_DESCRIPTION constant.

Built 6 new connectors:

IMAP Mailbox, POP3 Mailbox, SMTP Email — straightforward, no conditional fields.
Exchange Mailbox (IMAP), Exchange Mailbox (POP3), Exchange Send — each uses form.watch("MAIL_TRANSPORT") to conditionally show/hide the Graph-vs-protocol-specific fields (IMAP_HOST/IMAP_PORT, POP3_HOST/POP3_PORT, SMTP_HOST/PORT/SECURITY/USERNAME, GRAPH_BASE_URL, SAVE_TO_SENT_ITEMS), matching the JSON's showWhen rules.
Registry: added all 6 icon imports (EXCHANGE_IMAP/POP3/SMTP.svg, IMAP/POP3/SMTP.svg — confirmed these exist under img, resolved via the same @/assets/img/... alias the other 13 already use) plus registry entries with their notices.

johbaxter:
@neelneelneel @themaherkhalil as a FYSA

image

I had ai fix up the merge cons from this stuff, it seemed to pick up all the key changes and new forms for functions i spot checked a few.  But letting yall know.

PR #3615, opened 2026-08-31
SDK room changes

johbaxter:
## Description
- Removes mentions of Playground
- Moves room.ts to stores
- Adds room to insight initialization
- Add rule to Agents.md to only create types files when necessary, prefer inline typing
- Move agent code into stores (updated playground and workbench-assitant)

NEEDS TESTING

---

Update playground and client to use new agent store implementations

## How to Test

2. Expected outcomes


johbaxter (commented):
What is this file why did it create it? Im assuming this has to do with 47 stuff why did it create a new file

